PR #14, opened 2019-07-29
Remove H5pyDeprecationWarning

ArturoMS13:
Remove the following warning:
H5pyDeprecationWarning: dataset.value has been deprecated. Use dataset[()] instead.  "Use dataset[()] instead.", H5pyDeprecationWarning)


fonsocarre:
Thanks Arturo! The test fails because it cannot find a branch of xbeam and uvlm with the same name. But it is ok.

PR #19, opened 2019-10-08
WriteVariablesTime output global beam variables and consistent out dir

ngoiz:
I have modified the printing of structural variables in `WriteVariablesTime`. Previously when outputting a global beam variable (such as the quaternion) since `num_indices=1`  nothing would be saved. I have fixed this and if a variable has `num_indices=1` then it is assumed that it is a  global beam variable and it is saved in a file without any node number.

I have also modified the output directory structure such that it is consistent with other postprocs.

### DIFF
--- a/sharpy/postproc/writevariablestime.py
+++ b/sharpy/postproc/writevariablestime.py
@@ -34,6 +34,10 @@ class WriteVariablesTime(BaseSolver):
     settings_default = dict()
     settings_description = dict()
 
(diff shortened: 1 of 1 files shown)


ngoiz:
Apologies. This behaviour is already captured through `FoR_variables` rather than `structural_variables`.

ngoiz:
Returned to original behaviour but changed output directory

PR #25, opened 2019-11-15
Updated function calls for latest numpy 1.17

ngoiz:
Updated deprecated function calls with latest in numpy 1.17


ngoiz:
@fonsocarre please update environment files with latest packages

PR #26, opened 2019-11-15
Improved modal solver

fonsocarre:
Modal matrices can now be calculated with a given time step solution calculated previously.

Names of modal output files now reflect this new possibility and are accordingly updated


fonsocarre:
This is quite a simple one. I'll merge it myself.

Issue #32, opened 2019-11-21
Default value not correct in documentation when it is a numpy type.


fonsocarre:
I have just realised that in the documentation, when there is a default value from numpy (for example, a simple `np.zeros((3,))`, the output in the table looks like: `<sphinx.ext.autodoc.importer._MockObject object at 0x7faa13d4d198>`. This can be seen in  https://ic-sharpy.readthedocs.io/en/master/includes/solvers/structural/NonLinearStatic.html for the `initial_position` parameter.

I guess we just need to check if the type of the default value is derived from `mock` or not. I'm not sure about how to write automatically something like `np.array([0.0, 0.0, 0.0])`, although I'd be happy with just a `np.array`.

ngoiz:
Correct, sphinx will mock numpys (otherwise RTD will not build) so any np.arrays defaults are mocked. However, since the type of the setting itself is `list(float)` I think it is better if the default value is left as a list of floats, since the settings parser will convert it into a `np.array` anyway.

I would prefer this option, changing all defaults that are `np.arrays` to `list(float)` as per the setting type.

Another issue is when we have a default `np.array` multiplied by a `float`. In that case sphinx raises an error because it cannot multiply the mock by the float and does not import the docs. I'm chasing these and replacing with equivalent `list(float)`.

What do you think? last time I checked there was no simple way of adding the `np.arrays`

fonsocarre:
I guess it is a bit like when we have integers and you need to use `.value`, just something to take into account. My main problem with it is this: Let's imagine that our setting is `initial_position`, with default value `[0.0, 0.0, 0.0]` (that is, a plain python list).
If we give no value in the settings, we use the default, so the type is actually a python `list`.
If we give a value in the settings, the type will be a numpy `np.array`.
And there is no way of knowing this beforehand.

Do I make sense? Maybe we can indicate the defaults for lists as a string and process them if needed just as if they were the input?

ngoiz:
True, I think then I jumped the gun. I though if there was no given setting it would process it and return a `np.array`, but indeed it copies it and returns the `list`.

> Maybe we can indicate the defaults for lists as a string and process them if needed just as if they were the input?

I would prefer that actually. I.e. process all `list(float)` as `np.array` (regardless of them being defaults or user-specified and maybe adding a type `list` that just leaves it as a list regardless of whether the contents are `int`, `float` etc.

PR #33, opened 2019-11-21
Revert "Improvements in documentation "

ngoiz:
Reverts ImperialCollegeLondon/sharpy#31

Changes implemented for default values in #32 certainly introduces problems as discussed in #32. 

Tests not passing due to `np.array` being `list` since the value is obtained from a default parameter


fonsocarre:
I'll approve this once we fix the documentation issue this lists.

fonsocarre:
There is a better way of fixing this. I close this PR so that we can delete the `revert-31-dev_doc` branch.

PR #35, opened 2019-11-24
Change position of --name argument in docs

petebachant:
As part of the review for https://github.com/openjournals/joss-reviews/issues/1885.

I am running the docker image in Windows, and needed to change the order of this argument to avoid getting an error from bash: `/bin/bash: --name: invalid option`.


fonsocarre:
Thanks Pete, I appreciate it, as we somehow do not have a Windows machine to test this things.

PR #44, opened 2019-11-27
Updated installation instructions

ngoiz:
Included explicit instructions to install the release and development builds from source.

I haven't made any particular references to `xbeam` or `UVLM` releases, only a checkout from master is required. If you are happy we leave it as is.

The other option would be to have release tags in those repos and include them in the instructions, as is done with SHARPy.

### DIFF
--- a/docs/source/content/installation.md
+++ b/docs/source/content/installation.md
@@ -100,6 +100,7 @@ required to run SHARPy:
 ### Set up the folder structure
 
 0. Create the folder that will contain SHARPy and the underlying aerodynamic and structural libraries and `cd` into it.
+We will refer to this as the working folder.
     ```bash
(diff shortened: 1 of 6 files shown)


ngoiz:
I'll make the following updates prior to merging:
- [ ] Include instructions for release builds of uvlm and xbeam
- [ ] Release tags in uvlm imperialcollegelondon/uvlm#3
- [ ] Release tags in xbeam imperialcollegelondon/xbeam#6

fonsocarre:
I think we can include here too the addition of tags for the Docker container, so that every version of SHARPy is compiled with the proper release of xbeam and uvlm.

- [ ] Build Docker with the corresponding version of uvlm and xbeam instead of the latest one.

Issue #48, opened 2019-11-29
Clean up SHARPy linear interface with UVLM


ngoiz:
SHARPy's linear modules interface with UVLM through `sharpy/linear/src/libuvlm.py` whereas the rest of the code's interface is done in `uvlmlib.py`. To avoid confusion and mix ups, unify all and make the linear part interface through the main `uvlmlib`.

- [x] Merge `libuvlm` into `uvlmlib`

In addition, in several files in the linear source code, there are references to the c libraries that should be unified/cleaned up, mainly in 

- [x] `assembly`: uvlm lib directory import
- [x] `lib_dbiot`: clean up + uvlm lib directory import
- [x] `surface`: uvlm lib directory import
- [x] Consolidate `src/linear/libalg` into `algebra`

ngoiz:
Will also explore removing the vortex cutoff and use the epsilon (as is done with the nonlinear UVLM)

PR #55, opened 2019-12-10
Restructuring of A Short Guide to SHARPy

ngoiz:
I have restructured the content in `A Short Guide to SHARPy` into individual sections such that we can more easily update the content (rather than having the html notebook)

Also moved all media to `_static` so it gets copied as is into the `build` folder.


docs/source/content/debug.rst
```diff
@@ -0,0 +1,87 @@
...
+
+    * Check at the beginning of the execution the path to the SHARPy folder.
+
+* Not running the correct branch of the code.
+
+    * You probably want to use `develop`. Again, check the first few lines of SHARPy output.
+
+* Very different (I'm talking orders of magnitude) stiffnesses between nodes or directions?
+
+* The UVLM requires a smaller vortex core cutoff?
```
fonsocarre: Can you remove this one? We don't have a vortex cutoff for the nonlinear UVLM anymore.
```suggestion
* Maybe the UVLM requires a smaller a smaller vortex core cutoff (only for linear UVLM simulations, as the nonlinear uses another vortex core model).
```

docs/source/includes/index.rst
```diff
@@ -1,6 +1,16 @@
 SHARPy Source Code
 ------------------
 
+The core SHARPy documentation is found herein.
+
+.. note::
+
+	The docs are still a work in progress and therefore, most functions/classes with which there is not much user interaction are not fully documented.We would appreciate any help by means of your contributing to our growing documentation!
```
fonsocarre:
```suggestion
	The docs are still a work in progress and therefore, most functions/classes with which there is not much user interaction are not fully documented. We would appreciate any help by means of your contributing to our growing documentation!
```

fonsocarre: This is a single missing space. I'm resolving this directly.

docs/source/content/debug.rst
fonsocarre: Speaking about it, how difficult would it be to adapt the linear UVLM to the epsilon core model instead of the cutoff, or at least having the option to choose?

ngoiz: > Speaking about it, how difficult would it be to adapt the linear UVLM to the epsilon core model instead of the cutoff, or at least having the option to choose?

Good idea, I'm including this in #48

PR #57, opened 2019-12-10
Small typo in README.md and updates to it

fonsocarre:
Guys, the capabilities section only makes reference to the nonlinear part of the code, and for aircraft.
Any chance we can put together a few more lines about your part of the code? I feel like it is mature enough, and already available through master.


ngoiz:
I'm including this PR in #55 since there are updates to the README there. 

I will also expand on the capabilities (probably linking to the newly created capabilities page in the docs)

PR #73, opened 2020-04-11
Rework of direct balancing and test

outoforderdev:
Rework includes the following:
1) 'balreal_direct_py' is changed to become correct. Previous solution was based on a false mathematical statement, resulting in incorrect balancing matrices T and Tinv. New formulation is consistent with both Antoulas [1] and Gawronski [2] mathematically, with the notation borrowed from the latter.
2) Documentation part of the code is adjusted to be consistent with the notation in SHARPy and Gawronski [2]. Previous version of documentation had a mistake, where T and Tinv were reversed in the balancing formula. Should be correct now.
Also note, that the present SHARPy notation is actually the inverse of Matlab or Antoulas [1] - T and Tinv are the other way around. Decided not to change that, since too much of the code can rely on current notation.
3) Balancing test is changed. It couldn't previously spot any real mistakes in balancing, independent of what balancing matrices T and Tinv are. Addition of grammian calculation actually allows to compare observability and controllability and check whether they are actually equal.

### DIFF
--- a/sharpy/rom/utils/librom.py
+++ b/sharpy/rom/utils/librom.py
@@ -46,9 +46,9 @@ def balreal_direct_py(A, B, C, DLTI=True, Schur=False, full_outputs=False):
     The balanced system is therefore of the form:
 
     .. math::
-        \mathbf{A_b} &= \mathbf{T\,A\,T^{-1}} \\
-        \mathbf{B_b} &= \mathbf{T\,B} \\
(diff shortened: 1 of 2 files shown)


ngoiz:
I assume this is like #74 but without the last commit. When you open a pull request you can still add commits until it is merged and they will appear in the pull request history. So don't worry if it isn't perfect first time, you can modify and update as many times as you want and there will be a track.

PR #74, opened 2020-04-11
Rework of direct balancing

outoforderdev:
Rework includes the following:
1) 'balreal_direct_py' is changed to become correct. Previous solution was based on a false mathematical statement, resulting in incorrect balancing matrices T and Tinv. New formulation is consistent with both Antoulas [1] and Gawronski [2] mathematically, with the notation borrowed from the latter.
2) Documentation part of the code is adjusted to be consistent with the notation in SHARPy and Gawronski [2]. Previous version of documentation had a mistake, where T and Tinv were reversed in the balancing formula. Should be correct now.
Also note, that the present SHARPy notation is actually the inverse of Matlab or Antoulas [1] - T and Tinv are the other way around. Decided not to change that, since too much of the code can rely on current notation.
3) Balancing test is changed. It couldn't previously spot any real mistakes in balancing, independent of what balancing matrices T and Tinv are. Addition of grammian calculation actually allows to compare observability and controllability and check whether they are actually equal.

Also, as compared to previous pull request (which I closed) - I fixed an issue of an undefined package when running the test.


sharpy/rom/utils/librom.py
```diff
@@ -123,24 +127,73 @@ def balreal_direct_py(A, B, C, DLTI=True, Schur=False, full_outputs=False):
...
     # avoid Cholevski - unstable
-    hsv_sq, Tinv = np.linalg.eig(np.dot(Wc, Wo))
-    T = np.linalg.inv(Tinv)
 
-    # sort
-    iisort = np.argsort(hsv_sq)[::-1]
-    hsv = np.sqrt(hsv_sq[iisort])
-    T = T[:, iisort]
-    Tinv = Tinv[iisort, :]
+    # Added 11/04/2020: Building T and Tinv using SVD:
```
ngoiz, line 131:
```suggestion
    # Building T and Tinv using SVD:
```

It isn't really necessary to add the date in the code when you add changes since it is all tracked by git.

sharpy/rom/utils/librom.py
```diff
@@ -123,24 +127,73 @@ def balreal_direct_py(A, B, C, DLTI=True, Schur=False, full_outputs=False):
...
+    # Find T and Tinv:
+    S = np.sqrt(np.diag(hsv))
+
+    # Please note, the notation below is swapped as compared to regular notation in the literature.
+    # This is a known feature of SHARPy, hence it is maintained throughout (including documentation).
+    Tinv = scalg.inv(S) @ U.T @ Qot
+    T = Qc @ Vt.T @ scalg.inv(S)
+
+    # Building T and Tinv using eigenvalues instead:
+    # Note: the method commented out below does not work!
```
ngoiz, line 158: If the commented code below does not work please remove it

ngoiz:
Also, I have changed the base branch to `develop`. We tend to merge everything that is code related to `develop` to ensure it then gets some sort of use and testing prior to becoming the release version (`master`).

outoforderdev:
Thanks for the feedback @ngoiz, I believe I've just addressed the issues you've raised. Please let me know if there's anything else you'd like me to do.

ngoiz:
Done, thanks @outoforderdev !

PR #79, opened 2020-04-14
Several new aerodynamic enhancements

### LINKED ISSUE #75: Include viscous drag force from airfoil properties
Viscous drag force can be included in the computations. It is interpolated from airfoil properties tables: angle of attack - CL - CD - CM included in the aero.h5 file.

To activate this correction, set the new setting in coupled solvers: "correct_forces_method" = 'polars'.

### LINKED ISSUE #76: Recover vortex core in UVLM
The vortex core option is recovered in UVLM to avoid very large values of induced velocity that are unphysical. They come from the singularties in the vortex field. 

It slightly increases the computational time. 
It requires the updating of the UVLM library.

### LINKED ISSUE #77: Vortex wake managed by SHARPy
Previously, UVLM defined the initial shape of the wake. Now it is managed by SHARPy. Consequently it is easier to define a wake shape closer to reality to make convergence faster.

It is done through generators in the "AerogridLoader" solver. Check the new variables:
- wake_shape_generator
- wake_shake_generator_input

Right now, it is intended to be backwards compatible. However, this backwards compatibility will be removed in the future for code simplicity. Get used to describe wake shapes. Check the test cases for further information.

It requires the updating of UVLM.

### LINKED ISSUE #78: Allow CFL != 1in shedding step
The usual discretisation and implementation of the wake convection equation requires that it complies with CFL = 1 in both the shedding and the convection stages. 

This requirement is not longer needed in the shedding stage due to a new implementation. It requires the activation of "cfl1" = False in the aerodynamic solvers. 

Notice that it is only valid for the shedding stage. The rest of the wake still requires to comply with the condition of CFL =1.

It requires the update of UVLM.

ArturoMS13:
This pull request includes several aerodynamic enhancements:
Fixes #75
Fixes #76
Fixes #77
Fixes #78

It also includes minor bug fixing and tests updating.
It requires updating UVLM


sharpy/solvers/prescribeduvlm.py
```diff
@@ -57,6 +57,10 @@ class PrescribedUvlm(BaseSolver):
     settings_default['postprocessors_settings'] = dict()
     settings_description['postprocessors_settings'] = 'Dictionary with the applicable settings for every ``psotprocessor``. Every ``postprocessor`` needs its entry, even if empty'
```
ngoiz, line 58:
```suggestion
    settings_description['postprocessors_settings'] = 'Dictionary with the applicable settings for every ``postprocessor``. Every ``postprocessor`` needs its entry, even if empty'
```

Small typo

sharpy/utils/correct_forces.py
```diff
@@ -0,0 +1,186 @@
...
+
+# @gen_dict_force_corrections
+def polars(data, aero_kstep, structural_kstep, struct_forces):
+    r"""
+    This function corrects the aerodynamic forces from UVLM based on the airfoil polars provided by the user in the aero.h5 file
+
+    These are the steps needed to correct the forces:
+    * The force coming from UVLM is divided into induced drag (parallel to the incoming flow velocity) and lift (the remaining force).
+    * The angle of attack is computed based on that lift force and the angle of zero lift computed form the airfoil polar and assuming a slope of :math:`2 \pi`
+    * The dreag force is computed based on the angle of attack and the polars provided by the user
```
ngoiz, line 89:
```suggestion
    These are the steps needed to correct the forces:
    
        * The force coming from UVLM is divided into induced drag (parallel to the incoming flow velocity) and lift (the remaining force).
        * The angle of attack is computed based on that lift force and the angle of zero lift computed form the airfoil polar and assuming a slope of :math:`2 \pi`
        * The dreag force is computed based on the angle of attack and the polars provided by the user
```
Added indentation to list (for sphinx not to complain)

sharpy/linear/src/lin_aeroelastic.py
```diff
@@ -46,7 +46,7 @@ class LinAeroEla():
         SS (scipy.signal): state space formulation (discrete or continuous time), as selected by the user
     """
 
-    def __init__(self, data, custom_settings_linear=None, uvlm_block=False):
+    def __init__(self, data, custom_settings_linear=None, uvlm_block=False, chosen_ts=None):
```
ngoiz, line 49: Thanks for this!

I'm going to add a setting to the `LinearAssembler` (if it is not already there) to use this as well. I will make a new commit before we merge

sharpy/linear/src/assembly.py
```diff
@@ -1174,6 +1174,37 @@ def wake_prop(Surfs, Surfs_star, use_sparse=False, sparse_format='lil'):
...
+
+    n_surf = len(dimensions)
+    assert len(dimensions_star) == n_surf, 'No. of wake and bound surfaces not matching!'
+
+    for ss in range(n_surf):
+
+        M, N, K = dimensions[ss]
+        M_star, N_star, K_star = dimensions_star[ss]
+        assert N_star == N, \
+            'Bound and wake surface do not have the same spanwise discretisation'
```
ngoiz, line 1206:
```suggestion
            'Bound and wake surface do not have the same spanwise discretisation, N_star = %g and N = %g' % (N_star, N)
```

I've learnt the hard way that this helps with debugging :)

sharpy/utils/correct_forces.py
ArturoMS13: Thanks, I tried a bunch of different bullets and I couldn't figure out why sphinx didn't show it as a list

ArturoMS13:
Cool, I included a small explanation of the wake generation in issue #77 , I updated all the test cases to be used as examples and I think it is backwards compatible. I can elaborate it further for the release note.

PR #82, opened 2020-04-22
Pazi wing added to flying_wing template

outoforderdev:
Pazi wing was added as two classes in flying_wing.py template.

To use, refer to the [Goland Wing example](https://ic-sharpy.readthedocs.io/en/latest/content/example_notebooks/linear_goland_flutter.html) and change 'GolandControlSurface' to 'PaziControlSurface'. 

Note: the template currently produces unstable systems. Potential reason is unconverged static solution due to deformation under gravity. Currently WIP


cases/templates/flying_wings.py
```diff
@@ -1104,6 +1104,272 @@ def update_mass_stiff(self):
...
+        base_stiffness = np.diag([ea, ga_oup, ga_inp, gj, ei_oup, ei_inp])
+
+        self.stiffness = np.zeros((1, 6, 6))
+        self.stiffness[0] = base_stiffness
+
+        m_unit = 0.589 # kg/m
+
+        # Lumped mass to confirm wing properties:
+        # Bending:
+        # 5 N load at the tip. Expected deflection of about 62.6 mm at the tip (no gravity)
```
ngoiz, line 1199: This looks like a good future verification test. Would you mind adding a bit more detail on how to achieve that result? What discretisation and angle of attack did you use to achieve this?

outoforderdev: @ngoiz, I've added some details on the tests. They are verification tests against a full 3D FEA analysis in Abaqus

PR #94, opened 2020-08-17
New SHARPy examples in the documentation

### LINKED ISSUE #89: Potential bug in xbeam and cbeam interfaces
In files:
cbeam3_interfaces.f90
xbeam_interfaces.f90

in some subroutines such as cbeam3_slv_nlndyn_step_python, arrays **stiffness_db** and **inv_stiffness_db** arrays allocated as 

real(c_double), intent(in):: stiffness_db( n_mass, 6 ,6 )
real(c_double), intent(in):: inv_stiffness_db(n_mass, 6 , 6)

 while in some others including generate_xbeam, they are allocated as
stifness_db( n_stiffness, 6 , 6)
inv_stiffness_db(n_stiffness, 6 ,6 )

where n_mass and n_stiffness are number of mass and stiffness (with the type integer == c_int ). 
As far as I undrestand, n_mass and n_stiffness can be differenet in general as we input them to the subroutine separately.  

I believe the correct form is  stiffness_db( n_stfiffness, 6, 6).

Note : The issue is not limited to cbeam3_slv_nlndyn_step_python

ArturoMS13:
This PR includes two new examples: one cantilever beam and one simple airfoil to ease the use of the generate_cases library and thus the learning curve of SHARPy.

It requires the update of the submodules and the conda environment (sphinx package).

Fixes #89


docs/source/content/faqs.md
```diff
@@ -0,0 +1,44 @@
...
+
+    _[A]_ Make sure you have loaded the SHARPy variables using the command `source <path_to_sharpy>/bin_sharpy_var.sh`.
+    
+* __[Q] When plotting the aerodynamic forces in Paraview from the UVLM, the forces at the boundary between two 
+surfaces (for instance at the wing root) appears halved. Is my simulation incorrect?__
+
+    _[A]_ This is most likely not an issue with your simulation. We have observed over time that Paraview actually only
+    plots the result from one of the surfaces, hence why it appears half of what it should be. If you extract the forces
+    without using Paraview using the 
+    [WriteVariablesTime](https://ic-sharpy.readthedocs.io/en/master/includes/postprocs/WriteVariablesTime.html) postprocessor you will get the correct result.
```
ngoiz, line 19: @ArturoMS13, I wrote this a while back, do you know if this is correct? Or do the surfaces actually have half the force at a shared boundary?

sharpy/utils/plotutils.py
```diff
@@ -1,6 +1,8 @@
 """Plotting utilities
 """
 import numpy as np
+from mpl_toolkits.mplot3d import axes3d
+import matplotlib.pyplot as plt
```
ngoiz, line 5: Since we do not include these in the environment file (and I believe it used to cause issues with Docker), could we include these in a `try: ... except ModuleNotFoundError:` clause. Could you put these under the `if not plotly`?

utils/environment_linux.yml
```diff
@@ -164,4 +163,6 @@ dependencies:
   - zipp=0.6.0=py_0
   - zlib=1.2.11=h7b6447c_3
   - zstd=1.3.7=h0b5b093_0
+  - pip:
+    - sphinx_rtd_theme==0.5.0rc1
```
ngoiz, line 167: Can we use a "non-`pip`" `sphinx_rtd_theme`? I believe it made things easier in the Travis/Docker world...

ArturoMS13: Done! I didn't know this was the reason why all the pip packages disappeared at some point

sharpy/utils/plotutils.py
ArturoMS13: Done, thanks

docs/source/content/faqs.md
ArturoMS13: I have always though this was a visuallisation problem. When two surfaces match perfectly at two points, UVLM gives you half of the force on each surface. If you plot those forces on paraview you get two arrows one on top of each other. Visually you see a force which has half of the expected module, however you have two forces. As you say, when those forces are mapped onto the structure you get both contributions and, thus, the problem disappears.

My aerodynamic surfaces do not usually match but if you have an example I can doublecheck.

ngoiz: Ok, so the actual force at the boundary is half and the issue is that Paraview doesn't add them up but just shows one.

Then I think my little snippet of text in the FAQs is correct

ngoiz:
Thanks @ArturoMS13! I just changed the title for the changelog.

Feel free to merge whenever you want!

PR #95, opened 2020-08-18
New approach to multibody computations

ArturoMS13:
The main change introduced by this PR is that the structural multibody computations are performed in local A FoR. In DynamicCoupled, the structure variables are kept in the local A FoR of each body (positions, velocities ...) and they are only transformed onto the global A FoR when postprocessing operations are required (plotting, savind data...).
This should not influence results but it improves stability and convergence.

Other minor changes have also been included.


sharpy/postproc/aerogridplot.py
```diff
@@ -106,8 +106,9 @@ def run(self, online=False):
         # TODO: Create a dictionary to plot any variable as in beamplot
         if not online:
             for self.ts in range(self.ts_max):
-                self.plot_body()
-                self.plot_wake()
+                if self.data.aero.timestep_info[self.ts] is not None:
```
ngoiz, line 109: When would we have `None`s in `aero.timestep_info`?

sharpy/postproc/writevariablestime.py
```diff
@@ -169,20 +169,37 @@ def initialise(self, data, custom_settings=None):
...
         if 'FoR_number' in self.settings:
             pass
         else:
             self.settings['FoR_number'] = np.array([0], dtype=int)
 
+        if self.data.structure.timestep_info[it].in_global_AFoR:
+            tstep = self.data.structure.timestep_info[it]
+        else:
+            tstep = self.data.structure.timestep_info[it].copy()
+            tstep.whole_structure_to_global_AFoR(self.data.structure)
```
ngoiz, line 193: Would you mind adding to the docs that any output from `WriteVariablesTime` is in `A` FoR?

sharpy/structure/utils/lagrangeconstraints.py
```diff
@@ -848,7 +848,7 @@ def initialise(self, MBdict_entry, ieq, print_info=True):
             # for k, v in MBdict_entry.items():
                 # cout.cout_wrap(k, v)
 
-        cout.cout_wrap("WARNING: do not use fully_constrained_node_FoR. It is outdated", 3)
+        cout.cout_wrap("WARNING: do not use fully_constrained_node_FoR. It is outdated. Definetly not working if 'body' has velocity", 3)
```
ngoiz, line 851:
```suggestion
        cout.cout_wrap("WARNING: do not use fully_constrained_node_FoR. It is outdated. Definitely not working if 'body' has non-zero velocity", 3)
```

Spelling ;)

ngoiz (approved):
Thanks for the excellent work @ArturoMS13!

Obviously, I'm no MB expert so I cannot comment much there. I had a genuine question on `None`s appearing in `aero.timestep_info` and a typo I caught.

In any case good to merge.

sharpy/structure/utils/lagrangeconstraints.py
ArturoMS13: Thanks!

sharpy/postproc/aerogridplot.py
ArturoMS13: When you use the cleanup timesteps postprocessor, the old time steps are set to None to save RAM memory. I really need that because I run out of memory quite quickly. 

If you use the postprocessors "online" (inside DynamicCoupled, for example) there is no problem because you just write the last one. However, if you use them "offline" (in the "flow" variable of SHARPy), you may try to write time steps that were previously cleaned by the "CleanUp" postprocessor.

sharpy/postproc/writevariablestime.py
ArturoMS13: This piece of code does not mean that the output is going to be on the A FoR. All the postprocessors output the same variables as in the single body cases. In the case of "WriteVariablesTime" the variables are output in the same FoR in which they are stored. For example, the positions are in A FoR but the forces are in B FoR.

The issue with the multibody case is that we use several A FoR (one per body) and these lines of code just move everything to the same A FoR (named "global_AFoR") to have a coherent output.

That said, it is true that I always need to think twice about which FoR are variables referred to and it is also a source of problems for new users. Thus, I might add it somewhere in the documentation so it is clear from now on.

sharpy/postproc/aerogridplot.py
ngoiz: Sounds good, thanks!

sharpy/postproc/writevariablestime.py
ngoiz: Ok, makes sense. Many thanks again!

PR #97, opened 2020-08-19
Support linearised all-moving control surfaces

ngoiz:
Minor update to support linearised all-moving control surfaces.

In addition, minor bug fix for surfaces located at T-junctions and for control surfaces located in different beams but with the same control input.

Note, the linear control surface assembly is done on a per-node basis, a feature improvement would be to rework this on a per-element basis.

Minor fixes to the `LinDynamicSim` solver (useful to visualise the linearised control surfaces)


sharpy/linear/assembler/lincontrolsurfacedeflector.py
```diff
@@ -79,6 +79,8 @@ def generate(self, linuvlm=None, tsaero0=None, tsstruct0=None, aero=None, struct
         Returns:
 
         """
+        # In hindsight, building this matrix iterating through structural node was a big mistake that
```
ArturoMS13, line 82: Can you add the description of the input arguments?

sharpy/linear/assembler/lincontrolsurfacedeflector.py
```diff
@@ -79,6 +79,8 @@ def generate(self, linuvlm=None, tsaero0=None, tsstruct0=None, aero=None, struct
...
+        # has led to very messy code. Would rework by element and in B frame
 
         if self.aero is not None:
```
ArturoMS13, line 85: If the class has been initialised but it is still given the optional input arguments, it will ignore them. What do you think about changing "if self.aero is not None" by " if aero is None"?
```suggestion
        if aero is None:
```

sharpy/linear/assembler/lincontrolsurfacedeflector.py
```diff
@@ -166,16 +168,28 @@ def generate(self, linuvlm=None, tsaero0=None, tsstruct0=None, aero=None, struct
                                 with_control_surface = True
```
ArturoMS13, line 168: I think you can simplify the code changing lines 117-122 with:

for global_node in range(structure.num_node):
	i_elem, i_local_node = structure.node_master_elem

ArturoMS13 (changes requested):
Thank you very much for the commit, moving control surfaces give a great flexibility to the solver. in the future it would be great to generate a library with different moving surfaces. It could be similar to the library of Lagrange constraints.

ngoiz:
Found a bug when having the same control surface (i.e. elevator) split in two beams with different `xb` directions. This is fixed in e02a3880d6f35b9cddf6f1237377e97dea24e392, however, prior to merging the following needs to be done:
- [ ] Test implementation in the Horten wing (elevators on the same `xb`)
- [ ] Implement a simple test script 

I'm marking the PR as draft until I finish this

sharpy/linear/assembler/lincontrolsurfacedeflector.py
ngoiz: Added the inline docstrings. As well, removed the optional arguments since they are not used at all.

sharpy/linear/assembler/lincontrolsurfacedeflector.py
ngoiz: For the moment I need to keep this as is, for the case where a single node is attached to multiple elements. 

However, I do agree that the current implementation is inefficient. It should change to be done on a per-element basis but for the moment this gets the job done ;)

sharpy/linear/assembler/lincontrolsurfacedeflector.py
ngoiz: Indeed

ngoiz:
I have tested the fix for control surfaces that deflect in the same direction for control surfaces on beams with different `xb`s with positive results.

I included a simple script as a skipped unittest that can be used to visually test changes

ArturoMS13 (approved):
Good to go, thanks for the changes and the code cleaning

PR #100, opened 2020-08-25
Linearisation of externally applied follower forces

ngoiz:
The linearisation of externally applied follower forces that are defined in the material frame of reference (B) results in extra stiffening terms that must be included in the stiffness matrix of the linearised beam system.

This is similar to what was done with aerodynamic and gravity forces (in these cases, they were defined in the inertial G frame), so this expands to the case of externally applied follower forces (such as thrust).

All the math is included in the docstring in `lingebm.FlexDynamic.linearise_applied_forces()`.

A future enhancement would be to create the model for a linearised thrust input.


ArturoMS13 (approved):
Thanks for the work and the excellent documentation!

PR #101, opened 2020-08-26
Major new aerodynamic features

ArturoMS13:
This pull request includes major new features in the aerodynamic modules of SHARPy and requires the update of the UVLM submodule:

- Possibility to include viscous drag in aerodynamic computations through tabulated data
- Possibility to choose independently the vortex cut off radius between solid surfaces and wakes
- Possibility to use CFL different from one in wakes. Different interpolation and filtering options are included to support this feature
- Minor improvements in the computation of turbulent winds through bts files:
     - Use the same velocity of the background flow for all the points along the chord
     - Do not interpolate the background velocity on the wake nodes for efficency
- Minor bug fixes
- Possibility to write the background velocity at any point in the flow field
- Remove steady helicoidal wake solver which is not needed anymore. Check ``tests/utils/test_generate_cases.py`` to configure helicoidal wakes
- The method "initialise" of postprocessors accepts a new optional argument "caller" that tells the postprocessor which solver is calling it to allow the postprocessor to gather information upstream.

I have fixed a bug in the computation of the tangent vector in the algebra.py file but I know this is usually a source of errors between. Can you put special attention in the review of this change?


ngoiz (commented):
This is going to be a phenomenal feature, many thanks Arturo. Looking forward to it in SHARPy v1.2!

ngoiz:
I have updated this PR with the latest UVLM commits in `develop`, so it should be good to merge whenever you are ready. 

Thanks again for the awesome feature.

ngoiz (approved):

PR #104, opened 2020-09-16
Sends simulation info to xplane via udp

wong-hl:
Script that takes a pickled simulation as input and sends the simulation information to XPlane via UDP


scripts/xplaneUDP.py
```diff
@@ -0,0 +1,457 @@
...
+import binascii
+import pickle
+import os
+import numpy as np
+
+import sharpy.utils.solver_interface as solver_interface
+import sharpy.postproc
+from sharpy.utils.algebra import quat2euler
+
+import matplotlib.pyplot as plt
```
ngoiz, line 17: Do you need matplolib? We do not include matplotlib in our environment that is used for testing (`environment_minimal`) so we tend to wrap any matplotlib import around something like:
```suggestion
try:
    import matplotlib.pyplot as plt
except ModuleNotFoundError:
    with_matplotlib = False
else:
   with_matplotlib = True
```

and then preceed any matplotlib calls you make with an `if with_matplotlib:` (or something along those lines)

scripts/xplaneUDP.py
```diff
@@ -0,0 +1,457 @@
...
+
+            Therefore, typical message looks like,
+            DREF0 + (4byte value) + dref_path + 0 + spaces to complte the 
+            whole message to 509 bytes. (N.B. + => append. Don't include
+            them in the actual message)
+
+            E.G. to switch on anti-ice swithc
+            DREF0 + (4byte value of 1) + sim/cockpit/switches/anti_ice_surf_heat_left + 
+            0 + spaces to complte to 509 bytes 
+        '''
```
ngoiz, line 220: Do you mind changing the format of this docstring to Google style restructuredtext? We have some examples in the [docs](https://ic-sharpy.readthedocs.io/en/master/content/contributing.html#documentation) (point 4). It just needs to have the correct headers and if I'm picky double quote marks `"` rather than single `'`.

The main point of all this is that if you have the docstring in the correct format it gets picked up by our documentation generator and you will have your function documented in a nice-looking webpage.

scripts/xplaneUDP.py
```diff
@@ -0,0 +1,457 @@
...
+            except socket.timeout:
+                raise XPlaneIpNotFound()
+
+        sock.close()
+        return self.BeaconData
+
+
+if __name__ == "__main__":
+    import argparse
+    import time
```
ngoiz, line 437: This could be a good place to put a readme on how to use this script (as a docstring)

ngoiz (changes requested):
This is looking really good, many thanks for your contribution!

The main thing we need is a little readme/tutorial/doc on how to use your new script! That way I can try it out and make sure everything is ready for the merge!

I've also added minor comments, mainly on the docstring format such that they get picked up by the documentation generator.

Many thanks and good work!

ngoiz (approved):

PR #106, opened 2020-09-24
Fix accuracy problem in UVLMLin

### LINKED ISSUE #105: Numerical error in the linear UVLM output matrices when vortex radius is too small
This raised in imperialcollegelondon/UVLM#6

Calculating the vortex radius squared as one should leads to numerical issues to the C and D linear UVLM matrices when the parameter `vortex_radius` in the linear UVLM solver is small enough. 

Discussion in imperialcollegelondon/UVLM#6

ArturoMS13:
Fixes #105 and issue number 6 in UVLM
It requires recompiling UVLM


ngoiz (approved):
Makes sense to me. Many thanks for the quick fix!

PR #107, opened 2020-09-29
Generator to modify structural mass at runtime

ngoiz:
Introduces runtime generators, in this case a generator to change lumped masses during runtime. This could be used to simulate payload drops or air-to-air refuelling!

Runtime generators may be included in the `DynamicCoupled` solver, and the available options for this generator are listed in its documentation in `generators/`


ngoiz:
At the moment is out-of-date with `develop` but since there are quite a few PRs in line I'll update once they are all in

ArturoMS13 (approved):
Thanks for the contribution!

PR #108, opened 2020-09-29
Include linear UVLM stiffening and damping terms in the UVLM D matrix

ngoiz:
The stiffening and damping terms that arise in the linearisation of the UVLM to structural degrees of freedom projection matrices due to a change in the frame of reference are now included as part of the UVLM linear system.

Previously, these terms would be added to the stiffness and damping matrices of the structural subsystem, since they appear as perturbations to the structural degrees of freedom. However, the terms themselves are the projection of the non-zero aerodynamic forces at the reference condition. 

Thus, and for better portability of the UVLM, it is best that these terms are included as part of the direct term of the linearised UVLM system once it is projected onto the structural degrees of freedom, since the D matrix effectively transfers the inputs in the structural DOFs to forces, again in the structural frame. This allows for a standalone linear UVLM system projected onto the structural DOFs to be exported from SHARPy, without any further modification needed once the system is assembled.


ArturoMS13 (approved):
Thanks for the contribution.

PR #111, opened 2020-10-14
[fix] Fixed bug in Gridbox class

sduess:
- Removed all ".value" in Gridbox class
- Raised error since coordinates and spacings are not stored in dict format


ArturoMS13 (approved):
Thanks for solving this bug!

PR #118, opened 2021-01-12
Support to save ROM projector matrices

ngoiz:
Small enhancement to `SaveData` postproc that adds a `save_rom` setting that saves the ROM state space as well as the reduced order bases used for reduction.


ArturoMS13 (commented):
Thank you for the contribution. It looks good for merging.

ArturoMS13 (approved):

PR #120, opened 2021-02-01
Rigid coupled solver

ArturoMS13:
This PR mainly includes a new interface for a rigid structural solver for free systems.

It also includes minor fixes to SHARPy 1.2 and the update to UVLM and xbeam libraries to master.


sharpy/postproc/savedata.py
```diff
@@ -112,24 +113,6 @@ def __init__(self):
...
-                                                   'ct_gamma_dot_list',
-                                                   'ct_gamma_list',
-                                                   'ct_gamma_star_list',
-                                                   'ct_normals_list',
-                                                   'ct_u_ext_list',
-                                                   'ct_u_ext_star_list',
-                                                   'ct_zeta_dot_list',
-                                                   'ct_zeta_list',
-                                                   'ct_zeta_star_list',
-                                                   'dynamic_input'])
```
ngoiz, line 132: Just to confirm, these are no longer necessary because when `ClassesToSave` in the `save_timestep()` method is used these are not included?

I assume that is the case given the usual `remove_cpointers` but just to make sure

ArturoMS13: I remove these lines of code because ``settings_default['skip_attr']`` is already defined in the class attributes. These lines are just redundant.

I created the ``save_timestep`` method for clarity.

ngoiz: The issue is that if you give a list of attributes to skip as a setting, the `settings_default` are not used and all of those "rubbish" attributes are included. 

So, actually, the original `SaveData` was incorrect, line 117 in the snippet above should be `self.**settings**['skip_attr'].append([<all these attributes>])`. That way, you can include your own values to `settings['skip_attr']` and those `ct_` attributes are included regardless.

ArturoMS13: You are right. We should use "extend" instead of "append" to avoid nesting lists.
Thanks

ngoiz: Ahhh indeed it is `extend` rather than `append`. Thanks for spotting that!

ngoiz (approved):
Thanks for the great work as always, @ArturoMS13!

PR #124, opened 2021-03-15
Update linear UVLM to account for CFL not equal to one in the wake convection

ArturoMS13:
The linear UVLM has been updated to account for cases in which the wake discretisation does not comply with CFL = 1


sharpy/linear/src/assembly.py
```diff
@@ -1155,26 +1156,43 @@ def dfunstdgamma_dot(Surfs):
...
+        not allocate memory until this is accessed
+
+    Args:
+        MS (MultiSurface): MultiSurface instance
+        use_sparse (bool (optional)): Use sparse matrices
+        sparse_format (str (optional)): Use either ``csc`` or ``lil`` format
+        settings (dict (optional)): Dictionary with aerodynamic settings containing:
+            cfl1 (bool): Defines if the wake shape complies with CFL=1
+            dt (float): time step
+            vel_gen: velocity generator
```
ngoiz, line 1176: I see that the call to `settings['vel_gen']` is commented out, is it required? Also, speaking of the commented code below in lines 1230+, would you be happy if we got rid of it (provided all works)?

ngoiz (commented):
Thanks for this great effort! A very good addition to the world of linear systems.

I'm going to add a test using the full linear state-space and with that we should be good to go

ngoiz:
Actually, I'm going to implement the new test on the new linear development branch since otherwise it becomes a bit of a mess. From what I've tried on my branch it works very well and the performance improvement is massive!

So from my side its all ready, just if you can address that commented code note above and then we can merge

ngoiz (approved):

sharpy/linear/src/assembly.py
ArturoMS13: You are right, I have just cleaned it up

PR #130, opened 2021-03-26
Output directory in the location specified in settings[SHARPy][log_folder]

ngoiz:
Over time we've ended with all post procs having their individual `folder` setting to specify their output directory. This has led to several inconsistencies (some would have `folder + / case_name/` included in the code and others not) which always brought trouble when writing input files.

Since the use of different output folders for different postprocs is seldom used, the SHARPy output directory is now consolidated in what is provided in the `settings['SHARPy']['log_folder']` which will default to `./output`. In this directory, a subsequent directory with the case name will be provided such that all output folders (`aero`, `beam`....) will be found in `settings['SHARPy']['output'] / settings['SHARPy']['case']`. The output folder is also now an attribute of the `PreSharpy` class and can be usually accessed as `data.output_folder`.

The `folder` setting for each postproc has been removed. Tests have been updated.


sduess:
Well done, Norberto! It is a valuable change for SHARPy. 
The only thing I have noticed is a missing `import os`  in one of the post-processors (postproc/beamloads.py). It causes an error. 
One other minor suggestion: Let us not forget about our examples cases (e.g. cases\coupled\simple_HALE) and update them as well.

ArturoMS13 (changes requested):
Hi, 

thank you very much for the effort to standarise this feature.

I have made some commits:
- I have removed settings['folder'] from some solvers that still had it
- I have changed self.dir to self.folders in some solvers for standarisation
- I have removed the definition of the folder variable in tests and cases
- I have removed old BeamCsvOutput

I didn't date to touch the optimiser so you are welcome to have  a look at it. Can you please doublecheck the changes I made to PickleData and SaveParametricCase? I am not sure about them.

Best regards, Arturo.

ArturoMS13 (approved):

PR #148, opened 2021-05-10
Simulation settings check - Unrecognised settings raise Error

ngoiz:
Big functionality check with SHARPy checking that the simulation settings are correct or else raising an error. This feature is implemented with 2 main objectives in mind:
1. Clean up simulation files with endless deprecated settings that where being copied and passed on from sim to sim
2. Catch typos in settings that would mean SHARPy ignores that setting. It would previously throw a warning but we know how easy those are to miss. This means bye bye typo-induced head scratching of why this isn't doing what it is supposed to do.

Since it has been a collective effort getting all of this updated, please could you all have a look (obviously at files you personally haven't modified :) ). Thanks for all the work getting all those legacy cases and examples up to date!


sduess (commented):
Nice work and thanks for the useful contribution to SHARPy! I have just added a minor change to exceptions.py to avoid some duplicate code. The rest looks good to me!

ArturoMS13 (approved):
Thanks for all the contributions!

PR #160, opened 2021-05-28
Support element variables in UDP output

ngoiz:
Very minor enhancement to support element structural variables to be output in a UDP message (i.e. the CRV can now be sent).

As written out in the documentation, element variables should be included in the variables `yaml`  file as:

```
        - name: 'psi'  # element variable like the CRV
          inout: 'out'
          var_type: 'node'
          position: 3  # node id
          index: 2  # index dimension
```

and the code will look for the relevant element and local node in which the value is stored since these variables are typically stored in `[n_elem, i_local_elem, i_index]` form.

OP Update:
Changed to input to be the same as for `pos` variables.


tests/io/generate_pazy_udpout.py
```diff
@@ -158,8 +158,7 @@ def generate_pazy_udp(u_inf, case_name, output_folder='/output/', cases_folder='
 
     ws.config['AeroForcesCalculator'] = {'write_text_file': 'on',
                                          'text_file_name': ws.case_name + '_aeroforces.csv',
-                                         'screen_output': 'on',
-                                         'unsteady': 'off'}
+                                         'screen_output': 'on'}
```
ArturoMS13, line 161: Do we need the "screen_output" : "on" for a test case?

tests/io/sample_udp_inout/generate_pazy_test_io_local.py
```diff
@@ -158,8 +151,7 @@ def generate_pazy(u_inf, case_name, output_folder='/output/', cases_folder='', *
 
     ws.config['AeroForcesCalculator'] = {'write_text_file': 'on',
                                          'text_file_name': ws.case_name + '_aeroforces.csv',
-                                         'screen_output': 'on',
-                                         'unsteady': 'off'}
+                                         'screen_output': 'on'}
```
ArturoMS13, line 154: Same as previous comment

ArturoMS13 (approved):
Thank you for the contribution. I just made a very minor comment.

PR #162, opened 2021-06-23
UDP-Inout change for multiply cs_surface_deflections and loads/strain

Eriklyy:
Changed the inout_variables for the UDP connection, in a way that its possible to have multply cs inputs. Also it is possible to send loads and strains now as an output. A simple hale aircraft with two cs_surfaces is used as example.


sharpy/io/inout_variables.py
```diff
@@ -58,8 +58,31 @@ def get_variable_value(self, data, timestep_index=-1):
...
+                try:
+                    #First get the dict postproc_cell and the try to find the variable in it.
+                    get_postproc_cell = getattr(data.structure.timestep_info[timestep_index], 'postproc_cell')
+                    variable = get_postproc_cell[self.name]
+                except (KeyError, AttributeError):
+                    msg = ('Node {} is neither in timestep_info nor in postproc_cell.'.format(self.node))
+                    logger.error(msg)
+                    raise IndexError(msg)
+
+            #Had to add this part for for_pos and for_vel since they are arrays.
```
ngoiz, line 74:
```suggestion
            # Needed for for_pos and for_vel since they are arrays.
```
Well caught, just editing the comment to make it a bit more impersonal :)

ngoiz (approved):
Great addition, many thanks @Eriklyy

ngoiz:
I'll test this tomorrow (we have lost our automated testing for the moment) and if it all passes I'll merge into develop.

Thanks again!

ngoiz:
Everything runs. Many thanks!

Issue #171, opened 2021-08-26
Why no pip support?


DavidAnderegg:
Hi all
thank you for open sourcing this.

I just recently installed SHARpy and it seems a bit odd to me that you don't have a `setup.py`file. As far as i understand, the whole stuff lives in the folder i cloned and is made available through the `sharpy_vars.sh`. Is there a specific reason you dont want to install it into the anaconda directory and make it available system wide without this script?

Maybe i misunderstood something? Could you clarify this?

If there is no specific reason, i would look into it.

ngoiz:
Hi David,

That's certainly a very valid point. I take it that we started off with this `sharpy_vars.sh` approach and never really thought about using pip or conda builds (and to be completely honest, it is not my area of expertise). What we like (or at least what I sometimes do) about the current approach is that we can quickly change between SHARPy builds when developing new features. In my case I have a directory where I have my simulation scripts and then may change between a release version of SHARPy and another when I'm trying out some new stuff.

If you think SHARPy would benefit from pip support please feel free to look into it, I guess that's the real beauty of open source!

Thanks,
-Norberto

ngoiz:
Maybe @ArturoMS13 or @fonsocarre also have some insight

DavidAnderegg:
Hi Norberto

Thank you for the answer. I would be happy to contribute in this way and will look into it.

What you are saying about different builds could then be accomplished with different anaconda environments. You could have a `sharpy` and a `sharpy_dev` environment for example. Additionally, there would be some sort of version control for the end user.

ngoiz:
Hi David,

It sounds like a feature SHARPy would definitely benefit having! If you have time and look at it we would greatly appreciate it.

Keep us posted. Many thanks!

PR #175, opened 2021-09-07
Dev pipsupport

DavidAnderegg:
### Overview
As promised in #171 this PR makes SHARPy installable via pip. After building and testing, you simply install it in the current anaconda environment by running ```pip install .```. This copies all necessary files into the environment and makes the command ```sharpy``` globally available (as long as the anaconda environment is activated).

If you are developing SHARPy, you can install it in development mode with this command: ```pip install -e .```. This creates only a link to the location of the SHARPy files and allows "live editing" of its source while still having all the other benefits. More about it [here](https://pip.pypa.io/en/stable/cli/pip_install/#editable-installs).

### Changes

- created ```setup.py```which handles the install procedure (you might want to make sure I entered the proper information)
- added ```__version__``` variable to ```sharpy.__init__.py```. This tells the setup the current version of the library and has to be changed manually before a new release
- moved the ```cases``` folder into ```sharpy/cases```. This way it becomes part of the ```sharpy``` package itself. I guess this is the most controversial change I am proposing. But it is a lot cleaner and makes life easier (the examples can be imported everywhere via ```sharpy.cases....```). One could argue that it should not be part of the library as it does not really provide functionality. But I have seen this in the past (for example [here](https://github.com/OpenMDAO/OpenMDAO/tree/master/openmdao/test_suite/test_examples)) and I believe the benefits are worth it. What do you think?
- Changed the examples to import cases from SHARPy (e.g ```from sharpy.cases.templates import ...```)
- Removed the ```bin``` directory and its content
- added the function ```sharpy_run()``` to ```sharpy_main.py``` which gets executed whenever you call ```sharpy``` in the console. It is basically the same as ```bin/sharpy``` was
- Updated the docs to remove mentions of ```sharpy_vars.sh```

### General Notes
- I have broken the example ```nonlinear_t-tail_HALE.ipynb``` because i moved the ```cases``` folder. Maybe you can fix it? All the other examples seem to work.
- [Please don't do this again](https://github.com/ImperialCollegeLondon/sharpy/blob/20f93ecdc8b67165bb7ad28f2140953ebef8cd62/sharpy/utils/ctypes_utils.py#L17-L23). This completely breaks expected python behavior and took me quite some time to figure out.

I am quite sure i broke something it did not think about. Please test it and let me know what you think. Feel free to commit to my branch.


ngoiz:
Hi David,

Thank you very much for taking the effort and adding this to SHARPy, it will be a great addition that will make things a lot easier. I'll go through it over the next few days and comment here or commit directly to the branch if I see anything.

Again, many thanks!

ngoiz (approved):
Finally got around to reviewing and updating this, apologies and many thanks for adding this feature, @DavidAnderegg!!

All looks good, I added minor things and restored a data file that was lost thanks to gitignore I assume. 

Branch updated with the latest development branch that includes the automated unittests as a Github action that are also passing.

PR #176, opened 2021-10-13
Collective blade pitch PID control of wind turbines

ArturoMS13:
Draft pull request

New functionalities:
- Possibility to apply twist before sweep when generating the aerodynamic surface (Aerogrid)
- Apply a "stride" variable to some postprocessors to avoid running them every time step and save time and memory.
- Possibility to skip writing and plotting the variables associated to the wake to reduce memory requirements.
- InitialAeroelasticLoader: prescribes aeroelastic variables along time steps. We used it to compute the UVLM forces along the flight of an aircraft. We precomputed the shape of the aircraft with an external solver
- NoStructural solver provides the possibility to run UVLM rigid simulations with the features implemented in DynamicCoupled.
- tearDown function in solvers.
- map_forces_on_struct option in staticuvlm. It maps the UVLM forces on the structure which was not done before if it is not run within a coupled solver
- Rigid bodies in multibody solutions

Improvements:
- Restart functionality: save solvers information
- Apply rotation velocity to the PolarAeroForces computation for steady wind turbine cases
- Update wind turbine generators to the latest version
- Include linear solution for the airfoil gust in the test with wake CFL not equal to one.

Changes:
- I have change the name we used to load the "utils.settings" module to "su" because it was not consistent and there were too many things called "settings" within the solvers.
- Simplified and standard call to solvers "run(self, **kwargs) and simplified syntax for the default values that some solvers have at the beginning of the run function.
- Extract_resultants is computed in datastructures instead on the structural solvers.


ACea15 (approved):
Long-time standing Pull request. Important development and improvements to SHARPy from our wind turbines expert Arturo. Let's finally merge it!

PR #185, opened 2022-02-23
Unittest for nonlinear dynamic solver

sduess:
I added a dynamic unittest to check for errors in SHARPy's dynamic coupled solver. The resulting vertical load and pitching moment at the wing root of the HALE model after some timesteps are compared with reference values produced with SHARPy version 1.3. I have tested the test's functionalities in SHARPy's current "develop" branch and also with a non-functional commit in my fuselage branch. Note that the HALE generation file will be updated as soon as Alvaro's model generation method is pushed and merged.

### DIFF
--- a/.github/workflows/sharpy_tests.yaml
+++ b/.github/workflows/sharpy_tests.yaml
@@ -33,6 +33,8 @@ jobs:
           export PATH="$HOME/miniconda/bin:$PATH"
           hash -r
           conda config --set always_yes yes --set changeps1 no
+          conda update --name base conda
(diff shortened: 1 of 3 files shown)


sduess:
I have also updated the workflow in sharpy_test.yaml for the build checks using Python, following the guidelines here https://stackoverflow.com/questions/51712693/packagenotinstallederror-package-is-not-installed-in-prefix for updating conda.

ngoiz (approved):
Many thanks, Stef! 

(and thanks for fixing the condo issue!)

PR #186, opened 2022-03-11
Fix the sparse matrix in balancing ROM

AntonioWR:
A is a sparse matrix in csr_matrix(sparse) format, can not be directly passed into functions used in scipy _solver.py 
Sparse matrices do not work well with Scipy (Version 1.7.3) in the following code, so A is transformed into a dense matrix here first.


ngoiz (approved):
Many thanks for spotting this @AntonioWR! I have added a check to do the `.todense()` operation only if the A matrix is not a `np.ndarray`, but otherwise all good.

Thanks!

PR #194, opened 2022-05-05
Fix mass matrix generation for lumped masses in case of several masses per node

sduess:
Hey all,
Here is a quick fix for creating the mass matrix for lumped masses in beam.py when multiple lumped masses are connected to the same node. The added plus sign was accidentally deleted during this commit 8957405e09e6deca26401206dcb899b46a71606a and the previous correct version is simply restored. Without the addition assignment, the next loaded lumped mass on the same node overwrites the previous defined lumped mass there.


ngoiz (approved):
Thanks, Stef, for catching this one! It would have been tricky to pinpoint this bug

Issue #193, opened 2022-04-29
Control surface deflection sign convention


ngoiz:
Hi all,

This is related to #192 but really is a topic that needs addressing on its own as it is something that we may want to change, although, if we do, will imply backwards changes to all models with symmetric control surfaces.

## TLDR

Control surfaces deflections are not well defined according to a sign convention, which may pose problems in the future for antisymmetric control surfaces (ailerons) and may suppose now a difference in the linear versus nonlinear implementation of the control surfaces. We either change behaviour or continue as is, knowing how it works (a bug becomes a feature if you leave it long enough).

## Background

Control surfaces are implemented in the [aerogrid.py](https://github.com/ImperialCollegeLondon/sharpy/blob/master/sharpy/aero/models/aerogrid.py) file as part of the grid assembly process. 

The UVLM grid is assembled in "strips", where a strip is a chord wise section. These strips are concatenated in a span wise manner resulting in surfaces. 

A strip is generated here: https://github.com/ImperialCollegeLondon/sharpy/blob/41b6cb699fe33167d66cbe0c5ede1bdfaaeb50c4/sharpy/aero/models/aerogrid.py#L482

The whole surface is generated here:
https://github.com/ImperialCollegeLondon/sharpy/blob/41b6cb699fe33167d66cbe0c5ede1bdfaaeb50c4/sharpy/aero/models/aerogrid.py#L205

In the process of generating the strip, control surfaces are done first, in the local B frame by rotating the relevant chord wise section about `xb = [1, 0, 0]`. Then, the section is rotated according to any desired twist or sweep.

## Current process

These strips are concatenated sequentially in increasing node order. However, any rotation on the strip to assemble the surface is performed AFTER the control surface is implemented. Therefore, the control surface is rotated about an axis different from the material frame.

Take, for example the following configuration, where the arrow indicates the local xb, which is aligned with the beam and positive in the increasing node number direction
```
4 -<- 3 -<- 0 ->- 1 ->- 2 
```
The grid strips are generated sequentially for each element. Since the control surfaces are implemented always rotating about the local `[1, 0, 0]` and these are appended, the resulting rotation axis of each CS upon assembly looks like:
```
4 ->- 3 ->- 0 ->- 1 -> 2
```
which is NOT coincident with xb.

## Current behaviour

[...]

Issue #192, opened 2022-04-28
Flap inputs in state-space model not working in certain wing configurations


ngoiz:
Issue in linear control surfaces depending on the geometrical alignment of the wing, due to a bug in the linearisation process that, mistakenly, used the wrong indices in `for_delta` variable. 

The original intent was to mimic the behaviour of the nonlinear control surfaces where, for a control surface installed on right and left wings with the same `control_surface_id`, the deflection was symmetric despite an `x_b` axis in opposite directions, which would thereby violate the "right hand rule" for simplicity in the definition of the inputs. This led to the incorrect use of the `for_delta` variable that, depending on the wing alignment, could lead to a matrix mapping the displacements of the control surface to the lattice grid full of zeros.

Fix in progress...

### Originally raised in https://github.com/ImperialCollegeLondon/sharpy/discussions/190

<div type='discussions-op-text'>

<sup>Originally posted by **PiH2000** April 26, 2022</sup>
Hello,

I am trying to extract a state-space model for a wing with a flap at the tip.
However, the 2 columns in the B and D matrices corresponding to the flap inputs (deflection & velocity), and only them, are full of zeros. I have tried many different things, such as changing the surface from static to dynamic, increasing its size etc. but I keep getting the same issue.

The code that I am using is based on the FlyingWing class template. I have the attached the solver settings file. I am using a M=5, N=52 discretisation and the assembled aeroelastic system, in modal coordinates, is of the form:

Aeroelastic system assembled:
Aerodynamic states: 5460
Structural states: 10
Total states: 5470
Inputs: 17
Outputs: 15

Is there any setting in particular that would allow the flap to be modelled in the state-space system? Am I missing something?

Thank you for your help.
Pierre

[settings_M5N52u20.solver.txt](https://github.com/ImperialCollegeLondon/sharpy/files/8565257/settings_M5N52u20.solver.txt)
</div>

PR #196, opened 2022-05-10
Enhanced linear solver

ngoiz:
Numerous additions and improvements to the SHARPy linear solver. A big PR, but hopefully it will be worth it

## New features

* State-space variable tracker. Keeps track of state-space variables (inputs, outputs and states), showing as the system is modified by new gains or projections, how the input and output channels evolve. This is implemented in the `sharpy.linear.src.libss.StateSpace` and `sharpy.linear.src.libss.Gain` classes, using the variable tracker features developed in `sharpy.linear.utils.ss_interface`. Recommended for all new developments using state-spaces and gains.

* Stability derivatives post-processor. Detail settings [here](https://github.com/ngoiz/hale-ders/blob/main/Delivery/01_StabilityDerivatives/01_StabilityDerivatives.ipynb)

* New linear gusts. Support for single, span-wise constant input (`LeadingEdge`). Spanwise varying gusts supported through `MultiLeadingEdge` gust. This is done using interpolation from gust "monitoring-stations", that keep track of the convection of the gust onto the lattice vertices. 

* Enhanced `LinDynamicSim` solver with easier input settings. May present a few issues but in general it is working, although linear time domain simulations are not the stronghold of SHARPy

* Can convert linear UVLM to continuous time and assemble the aeroelastic state-space in CT as well

## Other changes
* Enhancements to `libss`
* Change name of `sharpy.linear.src.libss.ss` to `sharpy.linear.src.libss.StateSpace`, this accounts for the majority of the file changes
* Change of `Modal` solver setting from `write_dat` to `save_data`
* Fixes issue with linear control surface deflection in certain configurations (see #192). Note, however, that the sign convention for control surface deflections differs between the nonlinear and nonlinear implementation (see #193). This is included in the relevant docstrings..
* Includes PR #188


rafapalacios (approved):

sharpy/linear/assembler/linearbeam.py
```diff
@@ -147,30 +155,35 @@ def initialise(self, data, custom_settings=None):
...
-        self.state_variables = LinearVector(state_db, self.sys_id)
+
+        if num_dof_rig == 0:
+            state_variable_list = [
+                VectorVariable('eta', size=num_dof_flex, index=0),
+            ]
+        else:
+            state_variable_list = [
+                VectorVariable('eta', size=num_dof_flex, index=0),
+                VectorVariable('V', size=3, index=1),
```
ACea15, line 174: Comment/explain these variables, especially V and W which are not common in the beam formulation

sharpy/linear/assembler/lineargustassembler.py
```diff
@@ -1,58 +1,551 @@
...
+            zeta_proj = np.zeros_like(zeta)
+            _, m, n = zeta.shape
+            for i_m in range(m):
+                for i_n in range(n):
+                    zeta_proj[:, i_m, i_n] = zeta[:, i_m, i_n].dot(self.u_ext_direction) * self.u_ext_direction
+            max_chord_surf.append(np.max(zeta_proj))
+            min_chord_surf.append(np.min(zeta_proj))
+        return min(min_chord_surf), max(max_chord_surf)
+
+    def assemble(self):
```
ACea15, line 106: It seems like nothing will work if the inherited class does not define the method assemble?? If that's the case just make the class abstract and declare the method as 
from abc import ABC, abstracmethod 
LinearGust(ABC):
...
@abstractmethod
def assemble(self):

sharpy/linear/src/libss.py
```diff
@@ -148,6 +156,89 @@ def states(self):
     def states(self, value):
         self._states = value
 
+    @property
```
ACea15, line 159: for outputs, states etc. you call the setter within the getter. Not sure how rigorous this is: if you were to do instance.states = x, but instance.A.shape[0] was a different number, when retrieving instance.states you would get A.shape[0] not what you set to x... 
So in this case I would remove the setter altogether and just do  self._states = self.A.shape[0] in the getter. 
Minor thing anyway

sharpy/linear/src/lingebm.py
```diff
@@ -14,6 +14,7 @@
 import sharpy.utils.cout_utils as cout
 import sharpy.structure.utils.modalutils as modalutils
 import warnings
+from sharpy.linear.utils.ss_interface import LinearVector, StateVariable, OutputVariable, InputVariable
 
 
 class FlexDynamic():
```
ACea15, line 20: How well the linearisation plays with the polars corrections in the aero? It should be all right since the linearised external forcing terms are taken from the .structure.steady_app_forces which includes those...

sharpy/aero/utils/utils.py
```diff
@@ -121,3 +121,55 @@ def span_chord(i_node_surf, zeta):
     dir_chord = algebra.unit_vector(dir_chord)
 
     return dir_span, span, dir_chord, chord
+
+
+def find_aerodynamic_solver(settings):
```
ACea15, line 126: This is amendable for a general solution:
solver_interface.dictionary_of_solvers  -> to give you the solvers an loop over them 
if solver_i.solver_classification == 'Aero': 
    ... (same structure as in the code)
elif solver_i.solver_classification == 'Coupled: 
    ...

Worthy? Only if more solvers are to be added in the equilibrium solution. But it shows off the nice structure of the solvers class!!

docs/source/includes/cases/templates/template_wt/index.rst
```diff
@@ -0,0 +1,19 @@
+template_wt
++++++++++++
+
+template_wt
+
+Functions needed to generate a wind turbines
+
+Notes:
+    To load this library: import cases.templates.template_wt as template_wt
```
ACea15, line 9: import sharpy.cases.templates.template_wt as template_wt

sharpy/postproc/stabilityderivatives.py
```diff
@@ -0,0 +1,215 @@
...
+
+            target_system_derivatives.dict_of_derivatives['force_cs'] = target_system_derivatives.new_derivative(
+                'body',
+                'control_surface_derivatives')
+            target_system_derivatives.save(self.folder)
+            target_system_derivatives.savetxt(self.folder)
+
+        return self.data
+
+    def get_freestream_velocity(self):
```
ACea15, line 163: shouldn't be the same one that u_inf in settings?

ACea15 (approved):
Very good stuff, see if some of the comments are worth a quick update and let's merge!

sharpy/linear/assembler/lineargustassembler.py
ngoiz: Good point!

sharpy/linear/src/lingebm.py
ngoiz: A few things are missing, yes, the steady forces with the corrections are included but these only appear in the stiffening and damping terms added to the UVLM (i.e. geometric effects only). The UVLM matrices would be missing the updated-with-polar circulation at the ref state.

The other correction is done on the output of the aeroelastic system, with the polars being modified in stability axes to change the lift only at this stage. I just realised I have this elsewhere so I'll merge it in a separate PR in a second.

This is Sec 4.1.1 and Ch8 further work of my thesis :)

docs/source/includes/cases/templates/template_wt/index.rst
ngoiz: Fixed, thanks

sharpy/postproc/stabilityderivatives.py
ngoiz: Yes, but a few quirks are needed in the case of restarted solutions, because the settings may have not been parsed and thus are still strings

PR #200, opened 2022-06-15
Approximate linear lift curve slope with polar corrections

ngoiz:
Supports corrections from polar airfoil data on the linearised aeroelastic system. 

Modifications are only made to the lift force at the moment. The forces from the aeroelastic system are projected onto the local stability axis (drag, lift, etc) and a coefficient is applied based on the derivative of the polar data at the linearisation condition. The forces are then projected back onto the original frame.

Full details found in the docstring:
https://github.com/ImperialCollegeLondon/sharpy/blob/eceff9c70d77e100c358b1b9309eb88363cdebee/sharpy/generators/polaraeroforces.py#L219

@rafapalacios @ACea15 please can you have a look whenever you have a moment? maybe we don't want to merge this just yet or maybe you think its a valuable feature


sharpy/generators/polaraeroforces.py
```diff
@@ -213,10 +216,197 @@ def generate(self, **params):
...
+        forces_at_ref_gain = np.zeros_like(polar_gain)
+
+        jj = 0  # Global DOF index
+        for inode in range(nnode):
+            if aero_dict['aero_node'][inode]:
+
+                ### detect bc at node (and no. of dofs)
+                bc_here = structure.boundary_conditions[inode]
+
+                if bc_here == 1:  # clamp (only rigid-body)
```
ACea15, line 318: On a trim analysis, this very first node with bc=1 would still have DoF, therefore is this routine only valid  for A-frame node clamped conditions? It is fine if that's the case but we should remark it at the function docstring.

ACea15 (approved):
Thanks Norberto, just a couple of comments and it is good to go into develop. It is still a work in progress but it is self-contained and does not conflict with any part of the code, I believe, so let's merge to develop and keep working from there.   
I reckon it could be a very useful feature in production environments and it is not easy to navigate from the nonlinear to the linear part, to then apply sectional corrections on the state-space systems based on the derivatives... so nice work!!

sharpy/generators/polaraeroforces.py
ngoiz: The issue with the bc=1 node is that it doesn't have flexible DOF and hence the airforces at that node are not mapped to it but rather to the A frame (i.e. with the sum of all the forces). Therefore, to correct the aero force at the bc=1 node, I subtract all the other forces from the A frame, correct them, and sum them all up again:
https://github.com/ImperialCollegeLondon/sharpy/blob/684e5c30ff22a5068f867a255504c85dce7ba0b2/sharpy/generators/polaraeroforces.py#L385

sharpy/linear/assembler/linearaeroelastic.py
```diff
@@ -996,3 +1026,130 @@ def load_uvlm(filename):
         uvlm_ss_read = read_data
         return libss.StateSpace(uvlm_ss_read.A, uvlm_ss_read.B, uvlm_ss_read.C, uvlm_ss_read.D, dt=uvlm_ss_read.dt)
 
+
+    def polar_gains(self, data, ss):
```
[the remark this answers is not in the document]
ngoiz: Thanks, indeed it can go

ngoiz:
Thanks, Alvaro, for the review. I'm seeing if can fix an issue with the tests (running fine from the individual directory but failing from the top-level) and I'll merge this

PR #217, opened 2022-10-20
Some Enhancements and Fixes for the Polar Correction, Wake Discretisation, and Gust Velocity Field Generator

sduess:
Dear all, 

This is a pull request including some enhancements that I added to the gust velocity field generator, polar corrections, and wake discretisation, as listed in more detail below.  

### Enhancements

- Faster interpolation processes within the gust velocity field generator and airfoil polar corrections 

The 1-D interpolation method used so far from NumPy has been replaced by Scipy’s interp1d method. Scipy’s method offers the advantage that instead of an interpolated value a reusable function is generated. As a result, most of the computational work for the interpolation is done only once, while with NumPy's 'interp' function the interpolation has to be obtained for each grid point at each timestep, resulting in high accumulated computation time. A quick change sped up my dynamic FLEXOP simulations with continuous gust inputs by more than  70 %.  

- Input setting for desired gust-components 

Enables the user to specify the direction of the gust. This is useful when for e.g. continuous gusts only the z-component is considered, or if a comparison with wind tunnel tests is obtained featuring vertically mounted wings.  

- Allowing several wake parameter inputs for each lifting surface 

This is useful when you need different discretisation parameters for each wake vortex grid within the wake discretization scheme or in general if e.g. the wake tail does not need to have as many streamwise panels as the wing does for convergence. To realize multiple inputs while retaining combability where only one parameter is used as a setting input, SHARPy’s setting handling had to be extended, allowing multiple setting input data types and accessing a default value. For the multiple setting input, a test has been added to the unittest. 

- Polar Corrections: 
     - Input setting to skip the polarforce correction for user-defined lifting surfaces. 
     - Enable a user-specified angle of attack (AoA) at zero lift input. Before, the AoA at zero lift has only been computed with a theoretical formula for thin surfaces. For the airfoil used for the FLEXOP wing, this estimated angle was far from the actual value leading to significant errors in the induced AoA computation for cambered airfoils. 

### Fix 
- Fixing wrong area calculation for structural nodes with shared panels from different surfaces:
For structural nodes whose force is mapped from multiple adjacent panels that do not share the same lifting surface (e.g. centre node of a wing with separately defined right and left wing surfaces), the wrong area is computed. The reason being is that so far, only the area of one of the panels is considered leading to significant overestimations of the induced AoA at this node. In the new version, a check of each structural node for several mapped lifting surface panels is obtained and all areas are included in the area calculation.


sharpy/generators/polaraeroforces.py
```diff
@@ -231,9 +241,34 @@ def generate(self, **params):
...
 
         return new_struct_forces
     
+    def write_induced_aoa_of_each_node(self,ts, folder,list_aoa_induced):
+        '''
+        Writes induced aoa of each node to txt file for each timestep. 
+
+        Args:
+            ts (int): simulation timestep 
+            folder (str): output folder to which indced aoa files shall be exported
```
ACea15: *induced*

sharpy/generators/polaraeroforces.py
```diff
@@ -263,16 +268,40 @@ def correct_surface_area_in_case_of_shared_surfaces(self, inode, struct2aero_map
...
+        
+    def check_for_special_cases(self, aerogrid):
+        '''
+        Checks if the outboard node is shared by multiple surfaces or single wingtip panel. 
+
+        Args:
+            aerogrid :class:`~sharpy.aero.models.AerogridLoader
+        '''
+        # check if outboard node of aerosurface
+        self.flag_shared_node_by_surfaces, self.flag_wingtip_node = np.zeros((self.n_node,1)), np.zeros((self.n_node,1))
```
ACea15: these flags are vectors, len=n_node...

sharpy/generators/polaraeroforces.py
```diff
@@ -263,16 +268,40 @@ def correct_surface_area_in_case_of_shared_surfaces(self, inode, struct2aero_map
...
+        for inode in range(self.n_node):
+            if aerogrid.aero_dict['aero_node'][inode]:
+                i_n = aerogrid.struct2aero_mapping[inode][0]['i_n']                
+                isurf = aerogrid.struct2aero_mapping[inode][0]['i_surf']
+                N = aerogrid.aero_dimensions[isurf, 1]
+                if i_n in [0, N]:
+                    if len(aerogrid.struct2aero_mapping[inode]) > 1:
+                        self.flag_shared_node_by_surfaces = 1
```
ACea15: ... but here they are defined as int, [inode] missing? If so check the use of the flags in other places

sharpy/solvers/staticcoupled.py
```diff
@@ -194,7 +195,8 @@ def run(self):
                     struct_forces = \
                         self.correct_forces_generator.generate(aero_kstep=self.data.aero.timestep_info[self.data.ts],
                                                                structural_kstep=self.data.structure.timestep_info[self.data.ts],
-                                                               struct_forces=struct_forces)
+                                                               struct_forces=struct_forces,
+                                                               ts=0)
                 self.data.aero.timestep_info[self.data.ts].aero_steady_forces_beam_dof = struct_forces
```
ACea15, line 200: similar changes will be needed in `staticcoupledrbm.py` which also uses the polars

sharpy/generators/gustvelocityfield.py
```diff
@@ -250,15 +249,17 @@ def initialise(self, in_dict):
         settings.to_custom_types(self.settings, self.settings_types, self.settings_default)
 
         self.file_info = np.loadtxt(self.settings['file'])
-
+        self.initialise_inteprolation_functions()
```
ACea15, line 252: *interpolation*

ACea15:
Good stuff Stef, minor comments to be discussed and we can merge to develop.

sharpy/generators/gustvelocityfield.py
sduess: Thanks for spotting! Fixed in https://github.com/ImperialCollegeLondon/sharpy/pull/217/commits/c75c9b4cbfad39f523c20a35016fcae24bd91c33

sharpy/solvers/staticcoupled.py
sduess: Thanks for spotting! Fixed in https://github.com/ImperialCollegeLondon/sharpy/pull/217/commits/fcd30c718e9fa58ee8063be62507c72165fabbad

sharpy/aero/utils/airfoilpolars.py
```diff
@@ -52,11 +54,15 @@ def initialise(self, table):
...
         self.aoa_cl0_deg = matches[iaoacl0]
 
-        self.cl_interp = interp1d(self.table[:, 0], self.table[:, 1])
-        self.cd_interp = interp1d(self.table[:, 0], self.table[:, 2])
-        self.cm_interp = interp1d(self.table[:, 0], self.table[:, 3])
+        self.aoa_bounds = [np.min(self.table[:,0]),np.max(self.table[:,0])]
+
+        self.cl_interp = interp1d(self.table[:, 0], self.table[:, 1],bounds_error=False,fill_value=(self.table[0,1],self.table[-1,1]))
+        self.cd_interp = interp1d(self.table[:, 0], self.table[:, 2],bounds_error=False,fill_value=(self.table[0,2],self.table[-1,2]))
+        self.cm_interp = interp1d(self.table[:, 0], self.table[:, 3],bounds_error=False,fill_value=(self.table[0,3],self.table[-1,3]))
```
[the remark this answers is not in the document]
sduess: Good points Alvaro. The two different fill values given are for when not only the upper but also the lower end of the input polars are reached. I agree with you that it is easier to edit the polar input file directly to avoid user misunderstandings. I have removed the warnings and instead an error is raised by Scipy's interpolation function (see https://github.com/ImperialCollegeLondon/sharpy/pull/217/commits/737846bfe6f5ba2a33e7cc5e4f8d2dc55d0c83b9)

sharpy/generators/polaraeroforces.py
sduess: Thanks Alvaro. I have changed the function in a later commit, resulting in removing the description including the typo anyway.

ACea15:
Ready, thanks Steff!

PR #218, opened 2022-10-21
SHARPy Docker now releases to GitHub Packages 

ngoiz:
Release SHARPy Docker image automatically to GitHub Packages

This is done in a CI workflow through a new GitHub action. Master, release and old versions docker images of SHARPy should appear in the Packages page of the repo. The installation instructions have been amended to show how to download the image and build a container. The `Dockerfile` has been amended to install SHARPy through `pip`, as introduced in `v2.0`.

A minor update is the change of how we store SHARPy's version. I have added an argument such that you can query the code version through `sharpy --version` and stored the version in `sharpy/version.py` which should be the one version of truth and only file that needs modification once a new release is ready.


ngoiz:
Added some modifications such that the Docker image is only published if it is a push to `master`, `develop` or an annotated version tag starting as `v*`. For pull requests, it builds the image but does not published. Once this is merged into `develop`, the job `test_image_build` should be added as a required check to the currently protected branches

kccwing:
Hello @ngoiz , thanks for updating the docker image and the automatic github action to generate these! We aim plan to switch over to a much simpler environment.yml file as part of the upgrade from py3.7 to 3.10, in #222 from @ACea15 (environment_new.yml), and hence will merge #217 and #222, then this, then switching to master as the default branch, then a new version.

kccwing (approved):

PR #221, opened 2022-10-27
Handle absolute tolerance for structural convergence tests as input settings and increase default value

sduess:
The absolute tolerance used in xbeam for the convergence test is added as a user-defined input to the base structural solver and is passed to xbeam with xbopts. The main reason for the change is a bug in xbeam where the absolute tolerance was defined as too small to be able to catch significant small residuals in convergence tests for which the relative tolerance fails (see https://github.com/ImperialCollegeLondon/xbeam/pull/11). As a result, the structural solver did not converge for the external gust assembler test part of SHARPy's unittest. Further, it has been noticed that the absolute tolerance was defined differently in various functions.

Hence, this pull request includes the following changes:
- define the absolute tolerance for convergence checks within xbeam as a central parameter in xbopts
- increase the default value to catch very small residuals
- allows an adjustment of the absolute tolerance by the user if needed

### DIFF
--- a/sharpy/solvers/_basestructural.py
+++ b/sharpy/solvers/_basestructural.py
@@ -36,7 +36,11 @@ class _BaseStructural(BaseSolver):
 
     settings_types['min_delta'] = 'float'
     settings_default['min_delta'] = 1e-5
-    settings_description['min_delta'] = 'Structural solver tolerance'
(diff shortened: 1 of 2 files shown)


kccwing (approved):
Thanks @sduess!!

PR #223, opened 2022-11-11
Dev internals

ACea15:
Fix a couple of minor things such that:

- symbolic links are not listed in the __init__.py  that automatically loads solvers, generators etc. (requested by collaborators working in clusters). 
-  simulations don't break when there are autosaved (and the like) files created by editors.


kccwing (approved):
Thanks @ACea15! This should help with not crashing sharpy with temporary files mostly for developers.

PR #224, opened 2022-11-15
Pre-commit hooks and black code formatting

ngoiz:
Hi Team,

Another one from me...maybe this one is controversial 

I've added some standard pre-commit hooks to SHARPy in order to ensure code consistency and formatting, such that code reviews can be spent reviewing code and not formatting!

What are pre-commit checks? These are checks that run every time you commit your code and ensure that the formatting adheres to standard practice.

[Black](https://black.readthedocs.io/en/stable/index.html) is a formatter that will modify the `.py` files to ensure the format is consistent (i.e. line length, end of file etc.). Don't worry, it won't break the code as it checks that the python byte-content is the same before and after the operation.

As you can see, every single file in the repository has been modified to ensure consistent formatting. To avoid issues with `git blame` there is a new `.git-blame-revs-ignore` file that contains these massive commits so that the history of the file is unobstructed by these changes.

What is controversial? Black uses always double quotes for strings `"` rather than `'` so every string in the code has been modified to this practice. This can be of course reverted.

Black is only one stage of the process that formats the code. The next step would be using [flake8](https://flake8.pycqa.org/en/latest/index.html) that actually checks for various things before allowing a commit. These checks can be very strict (preventing commits with unused variables, unused import statements, etc.) which the tool cannot change automatically since it is modifying code. Therefore, when I tried using `flake8` a lot of modifications are needed in the code files for the checks to pass which could be a project on its own making the code compliant. It could be worth it in the long run though, we could start with many exclusions (errors you don't check for) and adding them step by step.

In any case let me know what you think about including these in SHARPy, additionally we'd just need to add a line to the CI pipeline to run these alongside the tests.

I have manually added `pre-commit` to the new environment file only. Not sure if you are looking to discard the others.

```
# See https://pre-commit.com for more information
# See https://pre-commit.com/hooks.html for more hooks
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v3.2.0
    hooks:
    -   id: trailing-whitespace
    -   id: end-of-file-fixer
    -   id: check-yaml
    -   id: check-added-large-files
-   repo: https://github.com/psf/black
    rev: 22.10.0
    hooks:
    -   id: black
```


sduess:
Hi Norberto,
Thanks a lot for still contributing and this proposal. I find your idea to include black very good. On the weekend, my brother wanted to introduce me to black randomly as well. I showed him your YAML file and he added some stuff, he thinks might be useful as well (including the controversial flake8). Also, he applied the suggested syntax from YAML of using two spaces for indentation instead of four. 
We will surely have a more detailed discussion in the next SHARPy team meeting!

ngoiz:
Hi Stef, 

Yes, I think it will be quite nice and preserve well the code in the long run with many contributors. It is going to be a bit of a hassle for everything to be compliant with flake8 as it doesn't let you commit with errors, but if it is done in chunks I think it is totally worth it.

Let me know how you decide to proceed in case I can offer a hand in reformatting parts of it!

rafapalacios:
Not going for this.

PR #234, opened 2023-05-15
Updates to Goland wing example

rafapalacios:
Minor fixes and a few more comments on the Goland wing flutter example


sduess (approved):
Thanks Rafa!

PR #239, opened 2023-05-31
New pip installation

ACea15:
Minor installation enhancement so that SHARPy can be installed automatically. Simply do  
```
    git clone --recursive git@github.com:ImperialCollegeLondon/sharpy.git
    cd sharpy 
    pip install . 
```
or `pip install -e .` for the developer mode.

- Passing the tests in MacOs, Ubuntu 22.04 and Centos 8.
- Remove yml files for unnecessary environments but left the latest one for reference (should be removed once we are sure this is working). I still put a conda environment in the root folder only with the tools required for building the Xbeam and UVLM. gfortran, eigen, lapack etc. If anyone has those in their system, then SHARPy can be installed on any environment or the python installation. But to inexperienced users asking about manually building those in their OS (package managers make it easy but I would not understimate some students when it comes to finding issues),  this environment should be the default. 
- The most important thing of this update is that Python packages required by the software are specified in the setup.py file in the field install_requires. This will make integrating SHARPy with other tools much easier.  
Extra packages are usually added for building the docs or for testing the code, since those are not needed by most users. If at some point testing is migrated to pytest (recommended), an entry would be added as "tests".  
For instance, to install the basic + docs packages, just do `pip install  .[docs]`. For the whole lot, `pip install  .[all]` 
- Automatic compilation of submodules when launching pip install: I didnt modify CMake so workflow and directories remain the same. Meaning one can still compile everything with `CMake .. & make install`  inside \build. 
    - I wanted to add options to the pip install as to be able to pass arguments to the compilation. But this is currently under discussion in the new pip: https://github.com/pypa/pip/issues/12010
One way to get around this in to use environment variables. So I added an option if someone absolutely don't want to do the compilation when doing pip install. Just set in the terminal `export PIP_NOBUILD=yes` prior to `pip install .` and it will install SHARPy without building the submodules. 
    -  I added an option, cause why not, for developers on the submodules to compile them from Python (here adding settings can be done easily). You just need to run `python setup.py build_subm`. And to add arguments, to compile in debug mode instead of Release, for example, just do:
    ` python setup.py build_subm --cmake-args="-DCMAKE_BUILD_TYPE=Debug"`
I have noticed compiling in Debug mode give quite a few warnings that could be fixed in the xbeam library. To keep in mind if someone wants to do some cleaning. Same for the CMakeLists.text.      

## To be done
- Test in your machines, let's see if  we can bring number of installation issues for other users to infinitesimal.
- Test Docker
- Update GitHub actions workflow
- Update the Docs and test building them. 

## For the future
- Cleaner setup.py file by moving settings to config file setup.cfg
- Use skbuild, as it the newer way to build low level libraries in the installation. Currently it fails on the Fortran build and it's under development.


ben-l-p (approved):
Installation with pip works on M2 Mac

rafapalacios:
After Ben's update of this pip, this pull request should not be needed anymore.

PR #241, opened 2023-06-27
Small additions to the installation guide.

boltyboi:
Hi,
As a completely new user of SHARPy I faced a few errors while installing the program. I made a few simple mistakes and so have changed a bit of the installation guide to help other newcomers not to make the same mistakes I did. In short, all I've done is change up and add small parts to one or two steps to make it clearer for newer users what to do and what not to, and a rough idea of how long certain steps might take.
Thanks.


rafapalacios (changes requested):
Many thanks for the comments. I've just asked for a minor change.

rafapalacios (approved):
Great. Thanks.

PR #242, opened 2023-06-29
Updates to plotutils.py and the cantilever_wing demo

boltyboi:
Edit: Wind turbine demo has also been changed, but I can't change the title.

**Plotutils:**
Added code for `custom_scaling` and `z_compression`. These are used in the cantilever wing demo to make the simulated wings appear more realistic. They are disabled by default and therefore shouldn't interfere with other demos that use plotutils.
Also added a little more detail describing what certain functions actually mean, and cleaned up the formatting a little bit in those areas.

**Cantilever wing demo:**
Fixed various typos and added extra clarifications on what parts of the code do. Mainly, I've added the `custom_scaling` and `z_compression` as toggleable features under the simulations, alongside a short explanation of what they do. This should hopefully make the simulations less confusing to newcomers if they decide to enable it. Both of these features are disabled by default and must be manually turned on.

**Wind turbine demo**
Fixed some typos and changed up the grammar in some areas, **the results I obtained are inaccurate, however, as when "Steady Simulation" is set to true the results seem to break. I'm not sure how to remove these from the commit though. Sorry about that.**


rafapalacios (approved):
Thanks for the fixes!

rafapalacios (approved):

kccwing:
Merging this - thanks Ozan @boltyboi!

PR #249, opened 2023-08-16
New Fuselage Model plus Minor Improvements

sduess:
Dear team, 

As promised, this is the big pull request mainly including methods to model nonlifting bodies such as fuselages with SHARPy.
  
**Nonlifting Body Model Capabilities** 
The main changes are in the UVLM: 
1. Adding Source Panel Method (SPM) to the aerodynamic code 
2. Coupling of the UVLM with the Source Panel Method to allow computations of fuselage-wing configurations and introduction of phantom panels for junction modelling. 
3. I introduced structs in the aerodynamic code (see struct_utils.h) to avoid page-long function inputs, it has also increased the speed of the code. 
In SHARPy directly, the following c  
1. Classes for the nonlifting body grid, its loader, and data structure (similar to lifting surfaces) have been added. I have used parent classes as all classes for lifting and nonlifting surfaces share many similarities. 
2. The SHARPy-UVLM interface as well as the solvers have been extended to be able to start solely nonlifting body computations but also steady and unsteady coupled simulations for example fuselage-wing simulations. 
3. Postprocessors (most importantly AeroforcesCalculator and AerogridPlot) have been extended as well. 
4. Fuselage-wing template model has been added. 
 
**Unittest** 
Unittest is successfully running and the following tests have been added to check the new capabilities: 

- Test for uncoupled SPM  
- Test for SPM coupled with UVLM. 
- Test for phantom panel methods. 

Further all dynamic unittests have been speed up for quicker checks using coarser discretisation, and smaller fsi tolerances and substeps. 

**Documentation** 
Besides the usual documentation of code, the SHARPy docs and Readme have been updated. For the UVLM, a code documentation has been started that covers 80 % of the code for now. 

**Additional Notes**
This branch has been successfully used to create plots for one of my journal papers as documented in this repository of mine  https://github.com/sduess/Enhanced_UVLM_nonlinear_aeroelastic.


sharpy/aero/models/nonliftingbodygrid.py
```diff
@@ -0,0 +1,116 @@
...
+                    matrix_nodes[:, idx, node_counter] = np.dot(Cab, matrix_nodes[:, idx, node_counter])
+            for dim in range(3):
+                matrix_nodes[dim, :, node_counter] += structure_tstep.pos[i_global_node,dim]
+
+                # get zeta dot in A frame (velocity due to pos_dot)
+                matrix_nodes_dot[dim, :, node_counter] += structure_tstep.pos_dot[i_global_node, dim]
+
+            for idx in range(numb_radial_nodes):
+                # get zeta dot in A frame (velocity due to psi_dot))
+                matrix_nodes_dot[:, idx, node_counter] += (np.dot(algebra.skew(omega_a), matrix_nodes[:, idx, node_counter]))
```
kccwing, line 92: I guess this bit is quite related to the original generate_zeta_timestep_info in aerogrid.py - in there I realised the rigid body velocity (linear and angular) for the most general expression for the strip velocities were missing (in eq 54 in screenshot attached). 

Specifically here, it seems that in line 92, matrix_nodes used was already inclusive of the rigid body position -> effectively it's the strip coordinates in G frame already (as processed in line 85) - but then we premultiply with another Cga afterwards in line 96 - possible typo?


Screenshot 2023-09-06 at 10 54 53 AM

sharpy/aero/utils/mapping.py
```diff
@@ -69,10 +70,11 @@ def aero2struct_force_mapping(aero_forces,
                 cbg = np.dot(cab.T, cag)
 
                 for i_m in range(n_m):
-                    chi_g = zeta[i_surf][:, i_m, i_n] - np.dot(cag.T, pos_def[i_global_node, :])
                     struct_forces[i_global_node, 0:3] += np.dot(cbg, aero_forces[i_surf][0:3, i_m, i_n])
                     struct_forces[i_global_node, 3:6] += np.dot(cbg, aero_forces[i_surf][3:6, i_m, i_n])
-                    struct_forces[i_global_node, 3:6] += np.dot(cbg, algebra.cross3(chi_g, aero_forces[i_surf][0:3, i_m, i_n]))
+                    if not skip_moments_generated_by_forces:
```
kccwing, line 75: sorry I'm not very familiar with this bit of the code - but why is this extra toggle for moments necessary?

sharpy/sharpy_main.py
kccwing, line 1: it seems like we have removed some lines for backwards compatibility - are they necessary for this PR to work?

sharpy/utils/generator_interface.py
```diff
@@ -41,7 +38,7 @@ def generator_list_from_path(cwd):
     onlyfiles = [f for f in os.listdir(cwd) if os.path.isfile(os.path.join(cwd, f))]
 
     for i_file in range(len(onlyfiles)):
-        if onlyfiles[i_file].split('.')[-1] == 'py': # support autosaved files in the folder
```
kccwing, line 44: I remember there being a discussion of the issue of temporary files (created when editing the code) - should we leave it as is, unless there's a specific reason I don't know?

kccwing (approved):
Thanks @sduess, great addition to the aero/fuselage modelling capabilities, very tidy and well-commented code!! Please find a few questions that I have as annotated - sorry if I am missing anything obvious!

sharpy/utils/generator_interface.py
sduess: Good spot! I just reversed this change

sharpy/aero/models/nonliftingbodygrid.py
sduess: Hey Kelvin,
let us talk Friday more about it. Even before this extension o, I think the rigid body velocities are added to zeta dot (calculated in *grid.py) within the UVLM.

sharpy/sharpy_main.py
sduess: Thanks for spotting! I reverted this file to its original version. I had to add some unused inputs  to some class functions (as done for many other solvers previously too) to make it work. I'll add it to my list of potential refactoring ideas.

sharpy/aero/utils/mapping.py
sduess: Somehow this moments introduce high unsymmetric forces in the fuselage beam. I could not figure the reason yet. Since we have only looked at rather stiff fuselages for now, the moment around a particular node should not change much. The (global) neutral point shift is mostly introduced by the longitudinal force distribution too. So, I basically neglect these local moments for the fuselage nodes. I added some comments in the code to have it better documented.

sharpy/aero/models/nonliftingbodygrid.py
sduess: Just to document the discussion we had here: The strip coordinates are only corrected by the beam coordinates expressed in the A-frame but not G-frame. So, the projection into the G-frame is valid.

PR #252, opened 2023-09-28
fix polars concatenation in assembly of aeroinformation - main

kccwing:
hotfix - requesting to merge to both main and develop.
issue with using overload of + operator - if polars of the same length, the result would be a summed array rather than concatenation


sduess (approved):
Thanks for spotting and fixing Kelvin!

PR #253, opened 2023-09-28
fix polars concatenation in assembly of aeroinformation - develop

kccwing:
hotfix - requesting to merge to both main and develop.
develop - together with small documentation updates
issue with using overload of + operator - if polars of the same length, the result would be a summed array rather than concatenation


sduess (approved):
Thanks for spotting and fixing this bug Kelvin!

PR #255, opened 2023-10-11
Update docker environment 

sduess:
The docker still uses the outdated minimal conda environment. This causes an error since it still uses Python 3.7 instead of 3.10.


kccwing (approved):
Thanks @sduess, good fix!

PR #257, opened 2023-10-11
fix [docker] use correct environment name in docker bashrc

sduess:
We had to adjust also the environment in the bashrc file to make the docker working from the start.


kccwing (approved):
Nice @sduess!!! Let's close this :))

PR #266, opened 2023-10-25
Changed Github runner from Conda to Mamba

ben-l-p:
### Changed Github runner from Conda to Mamba
The recent change to the new Conda environment involved swapping out minimal environment used by the Github runner to the regular environment. This causes the Ubuntu runner to fail due to running out of its 7GB of RAM, and so new SHARPy versions could not be verified.

To fix this, Conda has been swapped out with Mamba, which is a faster and less memory-intensive drop-in replacement. This should not have any effect on running SHARPy.

### Fixed Typos in Linear_Goland_Flutter.ipynb
Small errors in example notebook have been fixed, should not have any effect on the results.

### DIFF
--- a/.github/workflows/docker_build.yaml
+++ b/.github/workflows/docker_build.yaml
@@ -14,7 +14,7 @@ env:
 
 jobs:
   build-and-push-image:
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
     permissions:
       contents: read
       packages: write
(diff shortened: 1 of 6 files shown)


kccwing (approved):
Thanks! @ben-l-p

ben-l-p:
Moved all Github runner tests from using Conda to Mamba, a less memory-intensive alternative. Conda was found to require more than the 7gb of RAM available to the Ubuntu runner, causing tests to fail.

PR #271, opened 2023-12-06
Update XBeam version to include compiler optimisation

ben-l-p:
Updated XBeam to change compile flags used in GFortran. The `-ftree-parallelize-loops=4` optimisation flag was found to make the beam solver take considerably longer to run than without it. The `-ofast` flag was added as this should increase performance in some cases with no noticeable impact on results.


sduess (approved):
Hey Ben! Thanks a lot for spotting and speeding up the code.

Issue #208, opened 2022-07-28
Potential bug in /sharpy/structure/utils/modalutils.py


LeeElvis:
**Describe the bug**
Hi, there exists a small bug at line 289 in `/sharpy/structure/utils/modalutils.py`:
`for mode in range(num_rigid_body, NumLambda - num_rigid_body):` should be `for mode in range(num_rigid_body, NumLambda )`.
Thank you.

LeeElvis:
Hi, another potential bug was found in `sharpy/postproc/aeroforcescalculator.py`, the function of `file_output`: `force_matrix` and `moment_matrix`  are mismatched with the following code in that function, which may cause confusion in the output file 'aeroforces.txt'.
Thank you.

sduess:
Thanks @LeeElvis for spotting that. Both have recently been fixed in #273.

PR #273, opened 2023-12-18
A few minor bug fixes

sduess:
Hey,

Here are just a few minor bug fixes:

1. Fixed the headers and output shapes of the force and moment matrix that are written to a file in the postprocessor aeroforcescalculator.py ( #208).
2. There was a small bug in modalutils.py that caused not exporting 10 of the last mode shapes if rigid body motions are considered ( #208).
3. There's a small format bug that does not cause the Jupyter Notebook for the UDP control tutorial to open (at least for my visual code) and it's also not displayed in the documents (might the caused by the same format error).
4. Most importantly, I deleted some left-over debugging prints giving Ben headaches, although he hopefully uses the remove button himself next time (better than taking paracetamole).
5.  In one previous merge request, the indentation in the savedata.py postprocessor got confused and certain files were not outputted anymore for specific conditions. This is fixed and I updated some "remove if the file exists" commands for all output data cases, otherwise it would not overwrite previous data. 

Enjoy your christmas holidays!

### DIFF
--- a/docs/source/content/example_notebooks/UDP_control/tutorial_udp_control.ipynb
+++ b/docs/source/content/example_notebooks/UDP_control/tutorial_udp_control.ipynb
@@ -2368,9 +2368,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Run SHARPy Simulation"
-   "execution_count": 22,
-   ]
+    "### Run SHARPy Simulation"]
   },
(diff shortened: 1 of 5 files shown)


ben-l-p (approved):
Approved, changed look good and work for me. Thanks for removing the testing print statements, I can now rest easy :)

PR #275, opened 2024-01-19
Update noaero.py for consistency in function input

kccwing:
NoAero errors out because of missing optional input - fixed


sduess (approved):
Good spot Kelvin! We should really expand our tests and include all solvers to capture these things. Let's talk on Wednesday.

Issue #276, opened 2024-01-25
Scipy 1.12.0 Incompatible


SJ-Innovation:
**Describe the bug**
scipy.fftpack has been deprecated since 1.14.x, attempting to run with scipy 1.12.x results in import failure of "ifft" from scipy. 

**To Reproduce**
Clean build a Docker image runtime environment using current environment.yml which does not provide version compatibility information, attempt to run, ImportError. 

**Solutions**
Either add version compatibility information to environment.yml file, or update entire codebase to scipy 1.12.x 
PR for initial hotfix in progress.

kccwing:
Thanks @SJ-Innovation for spotting this! We will be addressing this in our next bug fix.

ben-l-p:
Issue should be fixed with the recent change to pip install [HERE](https://github.com/ImperialCollegeLondon/sharpy/pull/280), which updated the code to use the new FFT library. Any "recent" version of SciPy from the last few years should work fine.

PR #277, opened 2024-01-25
Add scipy version info to env yml

SJ-Innovation:
Fix for #276 
Include environment.yml scipy version information until rest of package is updated to latest scipy


kccwing:
Hello @SJ-Innovation, thanks for the patch and for the wonderful first contribution! We'll merge this but keep the issue open for now until we address updating scipy soon.

PR #279, opened 2024-01-27
Update beamplot.py to have stride option, consistent with aerogridplot.py

kccwing:
As title - small fix


kccwing:
Affected also by #276.

ben-l-p (approved):

PR #280, opened 2024-02-08
Added pip install (with docs)

ben-l-p:
Extension of Alvaro's pull request [HERE](https://github.com/ImperialCollegeLondon/sharpy/pull/239), with added documentation and changes to the Github workflows. The Docker build still uses Conda/Mamba - ideally I want to remove this dependency.

Also fixed SciPy issues, as it was using a depreciated FFT function


ben-l-p:
Hold off on approving - some packages for a install have been forgotten

ben-l-p:
Changes have been made to the Mayavi install (required for TVTK) as there is a bug which can causes installation to fail with pip. It now git clones the latest version for installation - this has been tested on Python 3.10/3.11 on both Ubuntu and Mac. 

Also added a second Conda environment for Apple Silicon, as Conda is yet to fully include GCC/GFortran for this. This is purely removing these packages so that it will still "work", you just need to install them yourself instead.

Should all be working now - please test!

rafapalacios (approved):
Working now. Very useful contribution. Thanks!

PR #281, opened 2024-02-26
Fixed crv2rotation

ben-l-p:
Found a bug in the conversion from a $\boldsymbol{\Psi}$ vector to a rotation matrix $\mathbf{R}$. This only applies for larger  $\Psi$ angles, as the small angle correction (which most "normal" cases will use) is correct. 

Intended expression:
$$\mathbf{R} =  \mathbf{I} + \sin||{\boldsymbol{\Psi}||} \tilde{\boldsymbol{\Psi}} + \frac{1-\cos{||\boldsymbol{\Psi}||}}{||\boldsymbol{\Psi}||^2}\tilde{\boldsymbol{\Psi}} \tilde{\boldsymbol{\Psi}}$$

Used expression:
$$\mathbf{R} =  \mathbf{I} + \frac{\sin||\boldsymbol{\Psi}||}{||\boldsymbol{\Psi}||} \tilde{\boldsymbol{\Psi}} + (1-\cos{||\boldsymbol{\Psi}||})\tilde{\boldsymbol{\Psi}} \tilde{\boldsymbol{\Psi}}$$

### DIFF
--- a/sharpy/utils/algebra.py
+++ b/sharpy/utils/algebra.py
@@ -562,8 +562,8 @@ def crv2rotation(psi):
         skew_normal = skew(normal)
 
         rot_matrix = np.eye(3)
-        rot_matrix += np.sin(norm_psi)*skew_normal
-        rot_matrix += (1.0 - np.cos(norm_psi))*np.dot(skew_normal, skew_normal)
+        rot_matrix += np.sin(norm_psi)/norm_psi*skew_normal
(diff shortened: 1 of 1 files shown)


kccwing (changes requested):
Thanks @ben-l-p! Note line 561 normalises the vector - then referring to eq C10 in Dynamics of Flexible Aircraft - the existing expression seems to be correct.

Screenshot 2024-02-26 at 20 37 12

ben-l-p:
I'm bad at maths, back to the drawing board...

PR #285, opened 2024-04-10
Fix Github Runner Docker build failing

ben-l-p:
Removed some unnecessary flags used when installing Mamba, causing it to fail. Ideally Mamba would be removed as the new pip install method does not require it, however CMAKE cannot find BLAS this way - tried a few install methods. This shouldn't cause any issues as long as the old environment in /utils/ remains.


kccwing (approved):
Thanks @ben-l-p for fixing this!!

PR #284, opened 2024-04-01
Update develop branch with main

ben-l-p:
Updating develop branch with the main branch. This includes the new pip installation method, as well as some other small fixes.

PR #288, opened 2024-04-25
Added ability to turn aligned grid off

ben-l-p:
Includes a new aerogrid setting 'initial_align' to perform flow alignment at the first timestep only, and use this same rotation for all time steps. It is also now possible to use no alignment at all - not recommended as this creates a different geometry as all rotations are with respect to the beam only.

### DIFF
--- a/sharpy/aero/models/aerogrid.py
+++ b/sharpy/aero/models/aerogrid.py
@@ -34,6 +34,8 @@ class Aerogrid(Grid):
 
         self.cs_generators = []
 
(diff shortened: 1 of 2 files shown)


ben-l-p:
Will require https://github.com/ImperialCollegeLondon/sharpy/pull/284 to be merged to pass tests

kccwing (approved):
Thanks @ben-l-p - let's come back to the rotation when issues come up!

PR #290, opened 2024-06-10
Updated docs to address issue with conda install failing

ben-l-p:
Some users have had an issue with Conda failing whilst creating the environment, giving the following message:
```bash
Collecting package metadata (repodata.json): - Killed
```
This is due to Conda running out of RAM during install, documented [here](https://stackoverflow.com/questions/74566580/why-is-conda-create-env-killed). This is in line with problems found a few months ago, with 8 GB of RAM often not being sufficient. Unfortunately Conda isn't very efficient with memory, and there is little we can do about this. 

 I have updated the docs to include 3 options to fix this issue:
  * Increase available RAM (if running on a compute cluster etc)
   * Use [Mamba](https://mamba.readthedocs.io/en/latest/), a more efficient drop-in replacement for Conda. This is what I personally use locally and find it to be vastly better — however, I can appreciate this isn't an option in some situations!
   * Create a blank conda environment and install the required packages:
     	```bash
      conda create --name sharpy python=3.10
      conda config –add channels conda-forge
      conda install eigen libopenblas libblas libcblas liblapack libgfortran libgcc libgfortran-ng
      ```

Thank you to the students who both found this issue and documented the third fix method!

### DIFF
--- a/docs/source/content/installation.md
+++ b/docs/source/content/installation.md
@@ -1,5 +1,5 @@
 # SHARPy v2.3 Installation Guide
-__Last revision 10 May 2024__
+__Last revision 10 June 2024__
 
 The following step by step tutorial will guide you through the installation process of SHARPy. This is the updated process valid from v2.3.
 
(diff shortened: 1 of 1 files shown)


rafapalacios (approved):
Thanks for the tips!

PR #297, opened 2024-08-29
Add hosting of built wheel at PyPI - main

kccwing:
Adding github actions to automatically release new wheel to PyPI on new version tag. Testing added for commits to prepare wheel sans upload. A few changes to setup.py necessitated to enable PyPI. Installation guide and github actions readme updated with (experimental!) documentation.

PR #298, opened 2024-08-29
Add hosting of built wheel at PyPI - develop

kccwing:
Identical to #297 - but merging into develop.


wong-hl (commented):
If it's identical to #297, can't you just merge the main branch into develop?

kccwing:
> If it's identical to #297, can't you just merge the main branch into develop?

usually I tend to merge main into develop right before a PR (from develop to main) - does it make more sense to do hot fixes merging one then the other, or does it not matter?

wong-hl:
> > If it's identical to #297, can't you just merge the main branch into develop?
> 
> usually I tend to merge main into develop right before a PR (from develop to main) - does it make more sense to do hot fixes merging one then the other, or does it not matter?

In terms of the end result, it doesn’t matter. The order only matters for the git history (mainly the merge commit). Ideally, the hot fix should’ve been merged into `dev` first, then merge `dev` into `main`. But, this is workflow stuff so it makes sense to merge it directly into `main`. 

This is really a style/preference sort of thing. By merging `main` into `dev`, you have one less PR and a cleaner git history as the same commits aren’t merged twice.

ben-l-p (approved):
Happy for this to be merged if the previous commits are squashed - good addition Kelvin!

kccwing:
Thanks both for the suggestions and reviews, @ben-l-p & @wong-hl!!

PR #299, opened 2024-09-10
Non-perpendicular hinge axis possible for multibody solver

kccwing:
Changes to hinge_for and hinge_node_for, and unit tests for multibody beam formulation using the augmented lagrangian method - cases with dominating scaling vs penalty factor checked, as well as a global rotation of the double pendulum with non-perpendicular axis. 3 new tests take in total of 3 minutes to run (3*30 time steps)

Second derivatives of constraint equations evaluated manually - to be replaced with AD version in due course.


sharpy/postproc/aerogridplot.py
```diff
@@ -199,7 +199,16 @@ def plot_body(self):
                     counter += 1
                     coords[counter, :] = aero_tstep.zeta[i_surf][:, i_m, i_n]
                     if self.settings['include_rbm']:
-                        coords[counter, :] += struct_tstep.for_pos[0:3]
+                        #TODO: fix for lack of g frame description in nonlineardynamicmultibody.py
+                        if struct_tstep.mb_dict is None:
+                            coords[counter, :] += struct_tstep.for_pos[0:3]
+                        else:
+                            #TODO: uncomment for dynamic trim
```
ben-l-p, line 206: Does this need to be uncommented for dynamic trim, or can this be removed?

sharpy/structure/utils/lagrangeconstraints.py
```diff
@@ -359,54 +360,903 @@ def equal_lin_vel_node_FoR(MB_tstep, MB_beam, FoR_body, node_body, node_number,
...
+            q[node_dof:node_dof+3] = node_dot_Ra
+            q[node_dof+3:node_dof+6] = psi_dot
+            # q[node_FoR_dof:node_FoR_dof+3] = node_FoR_va
+            # q[node_FoR_dof+3:node_FoR_dof+6] = node_FoR_wa
+
+            LM_Q[:sys_size] += penaltyFactor*np.dot(Bnh.T, np.dot(Bnh, q))
+
+            LM_C[:sys_size, :sys_size] += penaltyFactor*np.dot(Bnh.T, Bnh)
+
+        # # 16 canonical terms for (abcd)^T(abcd)
```
ben-l-p, line 839: Can we remove excess commented code?

sharpy/postproc/aerogridplot.py
kccwing: nope, these fix the issue that multibody simulations used to have outputs in the local A frame. The commented out portion will be useful for dynamic trim, but I haven't quite decided how best to proceed with it so I left the pointers there

tests/coupled/multibody/test_double_pendulum_slanted.py
```diff
@@ -0,0 +1,337 @@
+import numpy as np
+import unittest
+import os
+import shutil                 
+from sharpy.utils.constants import deg2rad
+
+class TestDoublePendulumSlanted(unittest.TestCase):
```
wong-hl, line 7: I know this is how we do unit tests in aerospace. But, this is not a unit test by the normal definition, as it should use [mocking](https://docs.python.org/3/library/unittest.mock.html) for the other components. This is technically an integration test. I'm ngl, I still haven't figured out what's an effective method for testing scientific code is 🤷‍♀️

ben-l-p: This is how other tests are defined, so even if there is a better way to do this, I would rather they be consistent for now. We can look into swapping them all to a better definition in a future release.

sharpy/structure/utils/lagrangeconstraints.py
```diff
@@ -359,54 +360,903 @@ def equal_lin_vel_node_FoR(MB_tstep, MB_beam, FoR_body, node_body, node_number,
...
+
+        #     LM_C[FoR_dof:FoR_dof+3, FoR_dof:FoR_dof+3] += penaltyFactor*mat                                                                          
+
+        # other LM_C derivatives for c dependencies in x1 and x2
+            # term 1-x1 - \frac{\partial}{\partial x_1}(a^Taq_1 + a^Tbq2 + a^Tcq3 + a^Tdq4) 
+            # da^Tdxaq_1 + a^Tdadxaq_1 + da^Tdxbq_2 + a^Tdbdxq_2 + da^Tdxcq_3 + a^Tdcdxq_3 + da^Tdxdq_4
+
+            mat = ag.multiply_matrices(-np.eye(3))
+                                       
+            vec = ag.multiply_matrices(-node_cga, node_dot_Ra)
```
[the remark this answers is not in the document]
ben-l-p: I am definitely a fan of the @ symbol for matrix multiplication - makes matrix algebra look much cleaner! Also, there's an identity matrix being multiplied by nothing here...

sharpy/structure/utils/lagrangeconstraints.py
kccwing: Addressed, thank you!

sharpy/structure/utils/lagrangeconstraints.py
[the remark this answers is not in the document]
kccwing: `mat` and `vec` structure is left intact for easier debugging and future work on these derivatives (until we replace the current approach) - open to replacing all with `@` - will look into this and do it at least for the code involved in this PR.

sharpy/utils/generate_cases.py
```diff
@@ -1979,7 +1979,7 @@ def check(self):
                 raise RuntimeError(("'%s' parameter required in '%s' lagrange constraint" % (param, self.behaviour)))
         has_behaviour = False
         for param, value in self.__dict__.items():
-            if not param in ['behaviour', 'scalingFactor', 'penaltyFactor']:
+            if not param in ['behaviour', 'scalingFactor', 'penaltyFactor', 'rot_axisA2']:
```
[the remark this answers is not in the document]
kccwing: `rot_axisA2` for backwards compatibility for existing job scripts, as well as consistency with `rot_axisB` in lagrangeconstraints.py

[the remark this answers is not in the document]
wong-hl: Ah the joys of consistently inconsistent naming conventions 😂

tests/coupled/multibody/test_double_pendulum_slanted.py
```diff
@@ -0,0 +1,337 @@
...
+        gc.clean_test_files(SimInfo.solvers['SHARPy']['route'], SimInfo.solvers['SHARPy']['case'])
+        SimInfo.generate_solver_file()
+        SimInfo.generate_dyn_file(numtimesteps)
+        beam1.generate_h5_files(SimInfo.solvers['SHARPy']['route'], SimInfo.solvers['SHARPy']['case'])
+        gc.generate_multibody_file(LC, MB,SimInfo.solvers['SHARPy']['route'], SimInfo.solvers['SHARPy']['case'])
+
+        # sharpy_output = sharpy.sharpy_main.main(['',
+        #                                      SimInfo.solvers['SHARPy']['route'] +
+        #                                      SimInfo.solvers['SHARPy']['case'] +
+        #                                      '.sharpy'])
```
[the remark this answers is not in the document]
kccwing: ideally i would create a class for the pendulum setup just like the [https://github.com/ImperialCollegeLondon/sharpy/blob/main/sharpy/cases/templates/flying_wings.py](flying wings) for the model build, then another higher level of abstraction for case setup like what @ACea15 had [https://github.com/ImperialCollegeLondon/sharpy/tree/dev_xbeam/sharpy/routines](done), but yes I do see value in this, especially the latter - I should have a good version of it by the time I am back in Feburary. Thanks!

tests/coupled/multibody/test_double_pendulum_slanted.py
kccwing: indeed @wong-hl - I came across this when I was looking for a way to have a parametrised `setup()` within uniitest - haven't heard about mocking so I'll read up on it!

ben-l-p (approved):
I cleaned up some of the code last week to try and make it a little bit less intimidating. I have been using this solver for a while already and have found it seemes to work as intended - again, great contribution Kelvin!

kccwing:
Appreciate the help with the PR @ben-l-p and @wong-hl! Thank you again :)

Shall I merge this one as well and then it goes in the latest (minor) version?

ben-l-p:
> Appreciate the help with the PR @ben-l-p and @wong-hl! Thank you again :)
> 
> Shall I merge this one as well and then it goes in the latest (minor) version?

I'm happy with it - if Hui Ling is too then we should be good to go. I would say we should leave this in develop now and merge to main in due course, as it shouldn't take me too long (don't quote me on this) to add my multibody edits too, so it can all be in one version.

kccwing:
Brilliant! No worries, and thanks again for the help and verification 😊

wong-hl (approved):
LGTM

PR #301, opened 2024-09-27
Fix AeroForcesCalculator giving forces in wrong frame of reference

ben-l-p:
This post-processor was previously giving erroneous results as the input forces were not in the expected (inertial) frame of reference, instead being in the flow direction relative to the body frame.

I did code up a more elegant post-processor, but it failed unit tests because of how the airfoil polars are implemented, and so I reverted to a modified version of the original. I really do not like this approach, as the forces used need to be mapped from B -> A -> G FoR, even though the UVLM forcing naturally should occur in the G FoR, however fixing this properly is not a trivial task and would require a rewrite of the polar code. This is something that should be investigated for anyone working with airfoil polars, as many of the aerodynamic forcing values within the solvers do not have the polar corrections applied, however it isn't made clear which do/don't.  

Also removed some excess commented code.

### DIFF
--- a/sharpy/postproc/aeroforcescalculator.py
+++ b/sharpy/postproc/aeroforcescalculator.py
@@ -5,15 +5,13 @@ import sharpy.utils.cout_utils as cout
 from sharpy.utils.solver_interface import solver, BaseSolver
 import sharpy.utils.settings as settings_utils
 import sharpy.utils.algebra as algebra
-import sharpy.aero.utils.mapping as mapping
-import warnings
 
 
 @solver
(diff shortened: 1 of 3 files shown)


rafapalacios (approved):
I have checked for the [dynamic gust](https://github.com/ImperialCollegeLondon/flexibleaircraftbook/blob/main/07UnsteadyAero/dynamic_gust.ipynb) and it has now only a very small difference with the results in the book, as expected. Thanks!

wong-hl (approved):
LGTM

PR #305, opened 2024-11-06
Add the automatic differentiation multibody solver based on JAX

ben-l-p:
I have added a new solver ```NonlinearDynamicMultibodyJAX``` with the similar functionality as ```NonlinearDynamicMultibody```, except it does not include any ```DynamicTrim``` routine (this can be added in due course when required). This new solver makes use of AD to calculate the Jacobians which leads to a much nicer and more compact definition of constraints, particularly for use of the penalty method. A lot of the key constraint types from the original solver are included here, such as:
 - Free hinge between beams with arbitrary axis line
 - Free hinge between beam and a fixed point in the inertial FoR
 - Spherical joint between beams
 - Spherical joints between beam and a fixed point in the inertial FoR
 - Fully constrained joint between beams
 - Fully constrained beam to a fixed point in the inertial FoR (this is effectively a less efficient way of created a “prescribed” beam, but was useful for testing)

The constraints numerics and the derivatives are defined in the ```sharpy/structure/utils/lagrangeconstraintsjax``` file. I have included new versions of all the other parts of the framework to prevent changes to one breaking the other solver etc., hence why ```sharpy/utils/multibodyjax``` and ```sharpy/solvers/timeintegratorsjax``` exist. 

Also included is the ability for controlled actuation between beams, which I primarily implemented for testing variable sweep wings, however the formulation/implementation is general for any 3D rotation. These can be controlled with the new ```sharpy/controllers/multibodycontroller``` controller type, which takes a Cartesian rotation vector time series as input. To allow this to sweep the wing correctly, the ability to warp the aero grid has been added with a new ```aerogrid_warp_factor``` parameter in the multibody file. This allows for a gradual sweep around a kink, and should not effect for existing cases, as it has no effect is the parameter is not included in the H5 file. 

A test case for this solver is included in the form of a flexible double pendulum comparison. A free double pendulum case is run, the angles from the two hinges extracted and applied onto a prescribed model, where both should yield the “same” result for structural deflections. 

Also included is some documentation on the multibody case files (should be general for both multibody solvers). In recent testing I have found a bug in the ```StaticTrim``` routine which was not present in v2.0 and I am currently investigating, which will be another PR in due course, but it is not connected to the multibody implementation. 

I have also done some code cleanups I found along the way, due to PyCharm largely doing it for me, but for files not related to the new solver the functionality should not have been changed. Lastly, I found one of the unit test cases was creating files which didn't get deleted at the end, which has also been fixed.

### DIFF
--- a/sharpy/aero/models/aerogrid.py
+++ b/sharpy/aero/models/aerogrid.py
@@ -193,11 +193,6 @@ class Aerogrid(Grid):
                 else:
                     global_node_in_surface[i_surf].append(i_global_node)
 
-                # master_elem, master_elem_node = beam.master[i_elem, i_local_node, :]
-                # if master_elem < 0:
-                    # master_elem = i_elem
-                    # master_elem_node = i_local_node
-
                 # find the i_surf and i_n data from the mapping
                 i_n = -1
                 ii_surf = -1
@@ -270,15 +265,28 @@ class Aerogrid(Grid):
                             raise NotImplementedError(str(self.data_dict['control_surface_type'][i_control_surface]) +
                                 ' control surfaces are not yet implemented')
 
-
+                # add sweep for aerogrid warping in constraint defintition
+                # if no constraint_xx.aerogrid warp factor is provided, this will be ignored
+                ang_warp = 0.
+                if structure_tstep.mb_dict is not None:
+                    if structure_tstep.mb_prescribed_dict is not None:
+                        for i_constraint in range(structure_tstep.mb_dict['num_constraints']):
+                            try:
+                                cst_name = f"constraint_{i_constraint:02d}"
+                                ctrl_id = structure_tstep.mb_dict[cst_name]['controller_id'].decode('UTF-8')
+                                f_warp = structure_tstep.mb_dict[cst_name]['aerogrid_warp_factor'][i_elem, i_local_node]
+                                ang_z = structure_tstep.mb_prescribed_dict[ctrl_id]['delta_psi'][2]
+                                ang_warp += f_warp * ang_z
+                            except KeyError:
+                                continue
 
                 node_info = dict()
                 node_info['i_node'] = i_global_node
                 node_info['i_local_node'] = i_local_node
-                node_info['chord'] = self.data_dict['chord'][i_elem, i_local_node]
+                node_info['chord'] = self.data_dict['chord'][i_elem, i_local_node] / np.cos(ang_warp)
                 node_info['eaxis'] = self.data_dict['elastic_axis'][i_elem, i_local_node]
                 node_info['twist'] = self.data_dict['twist'][i_elem, i_local_node]
-                node_info['sweep'] = self.data_dict['sweep'][i_elem, i_local_node]
+                node_info['sweep'] = self.data_dict['sweep'][i_elem, i_local_node] + ang_warp
                 node_info['M'] = self.dimensions[i_surf, 0]
                 node_info['M_distribution'] = self.data_dict['m_distribution'].decode('ascii')
                 node_info['airfoil'] = self.data_dict['airfoil_distribution'][i_elem, i_local_node]
--- a/sharpy/aero/models/aerogrid.py
+++ b/sharpy/aero/models/aerogrid.py
@@ -24,6 +24,7 @@ class Aerogrid(Grid):
     It is created by the solver :class:`sharpy.solvers.aerogridloader.AerogridLoader`
 
     """
+
     def __init__(self):
         super().__init__()
         self.dimensions_star = None
@@ -47,7 +48,7 @@ class Aerogrid(Grid):
                                          self.dimensions_star)
 
         # Initial panel orientation, used when aligned grid is off
-        self.initial_strip_z_rot = np.zeros([self.n_elem, 3]) 
+        self.initial_strip_z_rot = np.zeros([self.n_elem, 3])
         if not settings['aligned_grid'] and settings['initial_align']:
             for i_elem in range(self.n_elem):
                 for i_local_node in range(3):
@@ -62,7 +63,8 @@ class Aerogrid(Grid):
                 try:
                     self.airfoil_db[self.data_dict['airfoil_distribution'][i_elem, i_local_node]]
                 except KeyError:
-                    airfoil_coords = self.data_dict['airfoils'][str(self.data_dict['airfoil_distribution'][i_elem, i_local_node])]
+                    airfoil_coords = self.data_dict['airfoils'][
+                        str(self.data_dict['airfoil_distribution'][i_elem, i_local_node])]
                     self.airfoil_db[self.data_dict['airfoil_distribution'][i_elem, i_local_node]] = (
                         scipy.interpolate.interp1d(airfoil_coords[:, 0],
                                                    airfoil_coords[:, 1],
@@ -80,13 +82,12 @@ class Aerogrid(Grid):
         try:
             settings['control_surface_deflection']
         except KeyError:
-            settings.update({'control_surface_deflection': ['']*self.n_control_surfaces})
+            settings.update({'control_surface_deflection': [''] * self.n_control_surfaces})
 
         # pad ctrl surfaces dict with empty strings if not defined
         if len(settings['control_surface_deflection']) != self.n_control_surfaces:
-            undef_ctrl_sfcs = ['']*(self.n_control_surfaces - len(settings['control_surface_deflection']))
+            undef_ctrl_sfcs = [''] * (self.n_control_surfaces - len(settings['control_surface_deflection']))
             settings['control_surface_deflection'].extend(undef_ctrl_sfcs)
-
 
         # initialise generators
         with_error_initialising_cs = False
@@ -113,7 +114,6 @@ class Aerogrid(Grid):
         if with_error_initialising_cs:
             raise KeyError('Unable to locate settings for at least one control surface.')
 
-
         self.add_timestep()
         self.generate_mapping()
         self.generate_zeta(self.beam, self.aero_settings, ts)
@@ -136,13 +136,13 @@ class Aerogrid(Grid):
             cout.cout_wrap('     Wake %u, M=%u, N=%u' % (i_surf,
                                                          self.dimensions_star[i_surf, 0],
                                                          self.dimensions_star[i_surf, 1]))
-        cout.cout_wrap('  In total: %u bound panels' % (sum(self.dimensions[:, 0]*
+        cout.cout_wrap('  In total: %u bound panels' % (sum(self.dimensions[:, 0] *
                                                             self.dimensions[:, 1])))
-        cout.cout_wrap('  In total: %u wake panels' % (sum(self.dimensions_star[:, 0]*
+        cout.cout_wrap('  In total: %u wake panels' % (sum(self.dimensions_star[:, 0] *
                                                            self.dimensions_star[:, 1])))
-        cout.cout_wrap('  Total number of panels = %u' % (sum(self.dimensions[:, 0]*
+        cout.cout_wrap('  Total number of panels = %u' % (sum(self.dimensions[:, 0] *
                                                               self.dimensions[:, 1]) +
-                                                          sum(self.dimensions_star[:, 0]*
+                                                          sum(self.dimensions_star[:, 0] *
                                                               self.dimensions_star[:, 1])))
 
     def calculate_dimensions(self):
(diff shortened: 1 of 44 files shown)


sharpy/aero/models/aerogrid.py
```diff
@@ -270,15 +265,28 @@ def generate_zeta_timestep_info(self, structure_tstep, aero_tstep, beam, setting
                             raise NotImplementedError(str(self.data_dict['control_surface_type'][i_control_surface]) +
                                 ' control surfaces are not yet implemented')
 
-
+                # add sweep for aerogrid warping in constraint defintition
```
ben-l-p, line 268: New code for dynamically sweeping the aero grid. This shouldn't impact any existing cases as it will ignore it if the warp factor parameter does not exist.

sharpy/aero/utils/airfoilpolars.py
```diff
@@ -62,7 +61,7 @@ def get_coefs(self, aoa_deg):
         cd = self.cd_interp(aoa_deg)
         cm = self.cm_interp(aoa_deg)
 
-        return cl, cd, cm
+        return cl[0], cd[0], cm[0]
```
ben-l-p, line 64: Fixed one of the pesky errors converting an array to a scalar during the unit test - this seems to work fine.

sharpy/controllers/multibodycontroller.py
```diff
@@ -0,0 +1,180 @@
+import numpy as np
+import os
+
+import sharpy.utils.controller_interface as controller_interface
+import sharpy.utils.settings as settings
+
+
+@controller_interface.controller
+class MultibodyController(controller_interface.BaseController):
```
ben-l-p, line 9: New multibody controller for setting angle between beams.

sharpy/solvers/nonlineardynamicmultibodyjax.py
```diff
@@ -0,0 +1,368 @@
+import numpy as np
```
ben-l-p, line 1: New solver, very similar to the existing multibody solver in its inputs.

sharpy/solvers/steplinearuvlm.py
```diff
@@ -323,11 +312,10 @@ def run(self, **kwargs):
         """
 
         aero_tstep = settings_utils.set_value_or_default(kwargs, 'aero_step', self.data.aero.timestep_info[-1])
-        structure_tstep = settings_utils.set_value_or_default(kwargs, 'structural_step', self.data.structure.timestep_info[-1])
-        convect_wake = settings_utils.set_value_or_default(kwargs, 'convect_wake', False)
```
ben-l-p, line 327: I cut some parameters here as they aren't actually referenced again

sharpy/solvers/timeintegrators.py
```diff
@@ -71,6 +71,8 @@ class NewmarkBeta(_BaseTimeIntegrator):
 
     def __init__(self):
 
+        self.sys_size = None
```
ben-l-p, line 74: Some variables used later were not declared in __init__

sharpy/solvers/timeintegratorsjax.py
```diff
@@ -0,0 +1,209 @@
+import numpy as np
```
ben-l-p, line 1: New time integrator for the JAX solver, should have the same numerics as the existing implementation.

sharpy/structure/models/beam.py
```diff
@@ -544,7 +476,8 @@ def get_body(self, ibody):
         int_list_nodes = np.arange(0, ibody_beam.num_node, 1)
         for ielem in range(ibody_beam.num_elem):
             for inode_in_elem in range(ibody_beam.num_node_elem):
-                ibody_beam.connectivities[ielem, inode_in_elem] = int_list_nodes[ibody_nodes == ibody_beam.connectivities[ielem, inode_in_elem]]
+                ibody_beam.connectivities[ielem, inode_in_elem] = int_list_nodes[
```
ben-l-p, line 479: Another fix for converting arrays to scalars which creates a warning during unit test.

sharpy/structure/utils/lagrangeconstraintsjax.py
```diff
@@ -0,0 +1,498 @@
+from typing import Callable, Any, Optional, Type, cast
```
ben-l-p, line 1: This is where the multibody constraint numerics take place

sharpy/utils/generate_cases.py
```diff
@@ -1989,13 +1991,15 @@ def check(self):
             raise RuntimeError(("'behaviour' parameter is required in '%s' lagrange constraint" % self.behaviour))
 
 
-def generate_multibody_file(list_LagrangeConstraints, list_Bodies, route, case_name):
+def generate_multibody_file(list_LagrangeConstraints, list_Bodies, route, case_name, use_jax=False):
```
ben-l-p, line 1994: I need to add this additional setting as the generate case class checks for constraints in the existing solver; as I have new constraints which aren't implemented there, it fails when this is False.

sharpy/solvers/nonlineardynamicmultibodyjax.py
```diff
@@ -0,0 +1,368 @@
...
+import typing
```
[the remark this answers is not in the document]
ben-l-p: Typing is the built-in Python type hinting, so this doesn't require any new packages

sharpy/solvers/statictrim.py
```diff
@@ -317,17 +317,12 @@ def trim_algorithm(self):
 
     def evaluate(self, alpha, deflection_gamma, thrust):
         if not np.isfinite(alpha):
-            import pdb; pdb.set_trace()
+            raise ValueError
```
kccwing, line 320: Here I think the reason behind breaking and not terminating the run is because staticcoupled has no post-processing - so rather than killing it (and leaving no trace to what happened) they preferred to leave a break point here. If we haven't got plans to introduce post-processing for staticcoupled iterations, is there any neater alternative to just throwing an error here?

ben-l-p: I was having the trace cause me issues when running on the HPC a while ago, I'm not sure if it's good practice to have traces like this in “production” code (although I may be wrong). This error is thrown when one of the trim gradients is zero, which is currently occurring for some of my cases and I believe is a bug, that I'm currently looking into. If static coupled fails, the code won't get this far, as the FORTRAN will instead throw a singular matrix error.

sharpy/structure/models/beam.py
kccwing: good spot and fix!

tests/coupled/multibody/double_slanted_pendulum/test_double_pendulum_slanted.py
```diff
@@ -322,7 +301,7 @@ def test_doublependulum_hinge_slanted_lateralrot(self):
     def tearDown(self):
         solver_path = os.path.abspath(os.path.dirname(os.path.realpath(__file__)))
         solver_path += '/'
-        for name in [name_hinge_slanted, name_hinge_slanted_pen, name_hinge_slanted_lateralrot]:
```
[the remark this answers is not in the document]
kccwing: ahh sorry! my bad😂😂

tests/coupled/static/smith_g_2deg/generate_smith_g_2deg.py
```diff
@@ -208,7 +196,7 @@ def generate_aero_file():
 
     working_elem = 0
     working_node = 0
-    # right wing (surface 0, beam 0)
+
```
kccwing, line 199: my take is to leave these in - it helps with understanding the model generation procedure which if now there's no simplification coming up the pipeline seems to have a big learning curve.

tests/coupled/static/smith_g_4deg/generate_smith_g_4deg.py
```diff
@@ -207,7 +193,7 @@ def generate_aero_file():
 
     working_elem = 0
     working_node = 0
-    # right wing (surface 0, beam 0)
+
```
kccwing, line 196: likewise, I suggest leaving them in, happy to have a discussion

kccwing (approved):
Nice work, @ben-l-p! Thanks for building a parallel for most files used by the new jax routines - that should hopefully make the work of migrating over in the future much easier. Happy to merge given this passes all tests and that you've been using it day-to-day already.

sharpy/solvers/nonlineardynamicmultibodyjax.py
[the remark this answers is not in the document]
kccwing: ahhhhhh ok! thanks

sharpy/solvers/statictrim.py
kccwing: that makes sense - I meant more in the off chance that staticcoupled did return eg. same total forces for a parametrised run with different geometry (perhaps the staticcoupled get total forces function working on the timestep[-1] time information, and somehow got contaminated? just a wild guess) - that would kill the trim routine like what you've seen. but yeah I agree it is probably best to clean up production code for an HPC environment assuming no user input is possible further

wong-hl:
Btw I really like your PR message, it's super clear about what's in the PR etc

sharpy/aero/models/aerogrid.py
```diff
@@ -270,15 +265,28 @@ def generate_zeta_timestep_info(self, structure_tstep, aero_tstep, beam, setting
...
+                    if structure_tstep.mb_prescribed_dict is not None:
+                        for i_constraint in range(structure_tstep.mb_dict['num_constraints']):
+                            try:
+                                cst_name = f"constraint_{i_constraint:02d}"
+                                ctrl_id = structure_tstep.mb_dict[cst_name]['controller_id'].decode('UTF-8')
+                                f_warp = structure_tstep.mb_dict[cst_name]['aerogrid_warp_factor'][i_elem, i_local_node]
+                                ang_z = structure_tstep.mb_prescribed_dict[ctrl_id]['delta_psi'][2]
+                                ang_warp += f_warp * ang_z
+                            except KeyError:
+                                continue
```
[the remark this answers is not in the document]
ben-l-p: The constraints here are input as dictionaries and have different key-value pairs depending on their functionality (hinge axis, node number etc.). Warping the aerodynamic grid will occur here if a constraint has both ```controller_id``` and ```aerogrid_warp_factor``` entries, with the intended behavior to skip this code if a constraint is missing one. A key error will occur if it's a constraint that is missing either of these entries (and therefore is not a constraint which requires the warping). Of course, it could technically fail if I had coded it wrong and ```delta_psi``` is not defined, and this would incorrectly ignore this code, but I'm pretty sure that can't happen.

sharpy/controllers/multibodycontroller.py
```diff
@@ -0,0 +1,180 @@
...
+        "Write a time history of input, required input, " + "and control"
+    )
+
+    settings_table = settings.SettingsTable()
+    __doc__ += settings_table.generate(
+        settings_types, settings_default, settings_description, settings_options
+    )
+
+    def __init__(self):
+        self.in_dict = None
```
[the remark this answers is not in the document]
ben-l-p: I have included this to be consistent with the other controllers available, it's the settings dictionary, and I can add a comment to state this. I do not like how the controller code (for all of them) is structured, but I would rather they're all the same at least.

sharpy/controllers/multibodycontroller.py
```diff
@@ -0,0 +1,180 @@
...
+        if controlled_state["structural"].mb_prescribed_dict is None:
+            controlled_state["structural"].mb_prescribed_dict = dict()
+        controlled_state["structural"].mb_prescribed_dict[self.controller_id] = {
+            "psi": control_command,
+            "psi_dot": psi_dot,
+        }
+        controlled_state["structural"].mb_prescribed_dict[self.controller_id].update(
+            {"delta_psi": control_command - self.prescribed_ang_time_history[0, :]}
+        )
+
```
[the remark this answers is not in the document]
ben-l-p: Good spot!

sharpy/solvers/nonlineardynamicmultibodyjax.py
```diff
@@ -0,0 +1,368 @@
...
+                self.data.structure.timestep_info[-1].mb_FoR_vel[ibody, :] \
+                    = self.data.structure.timestep_info[-1].for_vel
+
+        # find total number of LC equations
+        dict_of_lc = lagrangeconstraints.DICT_OF_LC
+        lc_cls_list: list[typing.Type[lagrangeconstraints.Constraint]] = []
+        lc_settings: list[dict] = []
+        self.num_lm_tot = 0
+        for i in range(self.data.structure.ini_mb_dict['num_constraints']):
+            lc_settings.append(self.data.structure.ini_mb_dict[f'constraint_{i:02d}'])
```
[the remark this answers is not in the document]
ben-l-p: The number needs to be formatted to have two digits, i.e. 00, 01, 02. Not a fan of this method as it means this has to happen in a few places in the code, but unless if I overhaul the multibody then it's a necessity for backwards compatibility.

PR #306, opened 2024-11-07
Bug fix for generate_full_structure

kccwing:
not caught since nobody was using this - but will become a dependency of the statictrimMB solver so am getting this done now. thanks!


ben-l-p (approved):
Looks good - thanks for fixing Kelvin!

wong-hl (approved):

PR #309, opened 2024-12-09
Update body and wake FoR multiplication

kccwing:
silly mistake really - position in Gframe of multibody is offset by cga_0 not cga(t)

### DIFF
--- a/sharpy/postproc/aerogridplot.py
+++ b/sharpy/postproc/aerogridplot.py
@@ -208,7 +208,7 @@ class AerogridPlot(BaseSolver):
                             #     cga = algebra.euler2rot([0, self.data.trimmed_values[0], 0])
                             #     coords[counter, :] += np.dot(cga, struct_tstep.for_pos[0:3])
                             # except AttributeError:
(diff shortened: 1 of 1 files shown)


ben-l-p:
Apologies for taking so long to getting around to looking at this PR - this is an issue I was facing, so thank you for the fix. Should the time step you index not be the current one `self.data.structure.timestep_info[self.ts].cga()`, rather than the initial time step`self.data.structure.timestep_info[0].cga()`?

kccwing:
Thanks @ben-l-p, no worries! The trouble here is that the multibody solver, unlike the standard structural solver, stores in timestep_info['pos'] the displacements in the A frame, rather than in G - this is a patch to correct for that. Ideally this should have been done in the solver structure I agree - but the code uses the position too often so this is a "cleaner" patch.

kccwing:
ah sorry, I think I didn't answer your question - so basically the position is correct - it is just forgetting the initial FoR change.

ben-l-p:
Okay, that makes sense. Will approve it now, however will merge once we add a patch for the Mayavi issues, which I will investigate today

ben-l-p (approved):

PR #311, opened 2024-12-12
Mayavi github dependency restored for hotfix

kccwing:
https://github.com/enthought/mayavi/issues/1289

### DIFF
--- a/.github/workflows/sharpy_tests.yaml
+++ b/.github/workflows/sharpy_tests.yaml
@@ -39,8 +39,6 @@ jobs:
           git submodule init
           git submodule update
           git fetch --tags -f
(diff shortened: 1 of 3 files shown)


wong-hl:
I've fixed the `blas` bug and cleaned up the workflow file. It is currently failing on the unit tests highlighted in issue #313. I tried putting in the fix that @ben-l-p mentioned in the issue. However, the tarball can't be found making it fail at the build/installation stage. This is most likely due to them updating their default branch from `master` to `main` (see [mayavi PR](https://github.com/enthought/mayavi/pull/1330)). I tried updating the URL to have `main` instead of `master` and that failed too :(

PR #315, opened 2025-01-28
Added a plotting utility for aeroelastic modes

ben-l-p:
This includes a new  ```AeroelasticModal``` post processor which plots the aeroelastic modes, found from the linearised aeroelastic system. Whilst looking at the aeroelastic eigenvalues is useful for determining stability, this post processor is useful for investigating what the modes look like, which can be hard to interpret from a vector of complex numbers!

The output is similar to the ```Modal``` solver which plots the natural modes of the structure projected onto the aerodynamic grid as VTK files, however this here also includes the aerodynamic gamma terms for both the bound wing panels and the wake. As there is a phase component in the aeroelastic eigenvectors, I have included the ability to plot at evenly spaced points, which gives an animation of what the mode looks like. Please enjoy an example gif of the output below for the Pazy wing, however I have also validated it for the FLEXOP with more aerodynamic surfaces.

swept_pazy_flutter_cropped

To keep it compatible with the Krylov reduction (allowing for much faster analysis) I was required to make small changes to the linear system source code, only so that gain matrices are propagated to the post processor as well as information about the original linear system states. It can be very interesting to see what aerodynamic features the Krylov likes to keep, and what it discards. I increased the default Krylov order from 1 to 4, because 1 usually gives very poor results. 

I also removed some unused code here - the functionality of this solver should however be as before.

### DIFF
--- a/sharpy/postproc/aeroelasticmodeplot.py
+++ b/sharpy/postproc/aeroelasticmodeplot.py
@@ -15,8 +15,8 @@ class AeroelasticModal(BaseSolver):
     This class performs modal analysis on a linearised aeroelastic system. The modes are then plotted in phase steps,
     and the stability eigenvalues printed to the console.
 
-    Note: this does not have support for reduction on the state space model, and also requires a modal projection of the
-    structure.
+    Note: this supports model order reduction through the use of a Krylov projection. For this postprocessor to work,
+    the linearised model must use a modal projection of the structure.
     """
 
     solver_id = 'AeroelasticModal'
@@ -43,7 +43,7 @@ class AeroelasticModal(BaseSolver):
     settings_description['remove_conjugate'] = 'Remove conjugate modes'
 
     settings_types['max_rotation_deg'] = 'float'
-    settings_default['max_rotation_deg'] = 25.
+    settings_default['max_rotation_deg'] = 25.0
     settings_description['max_rotation_deg'] = 'Scale mode shape to have specified maximum rotation'
 
     settings_types['max_displacement'] = 'float'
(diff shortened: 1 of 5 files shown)


ben-l-p:
Okay, so I realise a similar, but more limited tool, already exists under ```AsymptoticStability```, may need to rethink

sharpy/postproc/aeroelasticmodeplot.py
```diff
@@ -0,0 +1,304 @@
...
+        if self.settings['remove_conjugate']:
+            first_eig_index = []
+            eig_set = set()
+
+            for i_eig, eig in enumerate(evals_c_ae):
+                # we here round the eigenvalues to 5 decimal places to prevent a pair not matching to machine precision
+                eig_pos = np.round(eig.real + 1j * np.abs(eig.imag), 5)
+                if eig_pos not in eig_set:
+                    eig_set.add(eig_pos)
+                    first_eig_index.append(i_eig)
```
wong-hl, line 107: Style suggestion: It might be worth renaming `i_eig` to `eig_index` (or something to that effect). As this deals with complex numbers, `i_eig` could be misconstrued as being the imaginary component. 

I don't know if this affects the logic of this, but would a dictionary be a better data structure as sets aren't ordered? That way, we'd be able to guarantee the index corresponds to a given `eig_pos`.

sharpy/postproc/aeroelasticmodeplot.py
```diff
@@ -0,0 +1,304 @@
...
+            self.data.structure.timestep_info[self.settings['use_custom_timestep']].modal['eigenvectors'].shape[1]
+
+        # add zeros to account for the clamped nodes which are not included in the structural eigenvectors
+        struct_evects = np.zeros((6 * num_node, num_modes_struct))
+        struct_evects[6:, :] = self.data.structure.timestep_info[self.settings['use_custom_timestep']].modal[
+            'eigenvectors']
+        node_disp_base = struct_evects @ evecs_ae_trunc[state_index['q'], :]
+
+        # mode shapes of the structural nodes, bound gamma and wake gamma
+        node_disp = np.einsum('ij,k->ijk', node_disp_base, phase).real  # [num_dof, num_lambda, num_phase]
```
wong-hl, line 141: Love the use of `np.einsum`!

sharpy/postproc/aeroelasticmodeplot.py
```diff
@@ -0,0 +1,304 @@
...
+                               np.einsum('ij,k->ijk', aero_evects[full_state_index['gamma_w'], :], phase))).real
+            pass
+        else:
+            raise KeyError("No aerodynamic states found in linear model")
+
+        # split structure into displacements and rotations
+        num_dof = node_disp.shape[0]
+        num_node = int(num_dof // 6)
+
+        i_pos = np.array(list(chain([6 * i, 6 * i + 1, 6 * i + 2] for i in range(num_node))))
```
wong-hl, line 167: Is the `chain()` equivalent to flattening the nested list?

sharpy/postproc/aeroelasticmodeplot.py
```diff
@@ -0,0 +1,304 @@
...
+        # add wake data
+        for i_wake, wake in enumerate(zeta_w):
+            zeta.append(np.tile(np.expand_dims(np.transpose(wake, (2, 1, 0)), axis=(3, 4)),
+                                (1, 1, 1, self.settings['num_modes'], self.settings['step_count'])))
+            zeta_delta.append(np.zeros_like(zeta[-1][:, :, 0, ...]))
+            is_bound.append(False)
+            surf_number.append(i_wake)
+
+        # create folder to save modes to
+        folder = self.data.output_folder + '/aeroelastic_modes/'
```
wong-hl, line 253: Does hard-coding the `/` in the path mean that SHARPy doesn't work on non-unix systems? If so, it might be worth progressively moving towards using `os.path.join()` [method](https://docs.python.org/3/library/os.path.html#os.path.join)

wong-hl (commented):
LGTM! Thanks for cleaning up the code in the other files too

sharpy/postproc/aeroelasticmodeplot.py
ben-l-p: Yes in this case it is. It is really best used to lazily evaluated generator expressions together, but it gives me the correct behaviour for lists so I'll use it

sharpy/postproc/aeroelasticmodeplot.py
ben-l-p: This is copied from the equivalent code in the modal solver, so have tried to keep things consistent. Can change this if anyone actually encounters issues from it, but if so the issue would need addressing elsewhere in the code too

sharpy/postproc/aeroelasticmodeplot.py
ben-l-p: Good spot. I think I was lucky in the fact that because I'm doing nothing crazy with the set, they were staying ordered - have swapped it over to a dictionary. Thanks for the pointers!

sharpy/postproc/aeroelasticmodeplot.py
wong-hl: Yeah, I think it would cause it to break in other parts of the code. It was a thinking aloud thing about if it would be worth having it such that newer code uses the `os.path.join()` and when we encounter the old style, to migrate it over. It boils down to the balance between consistent code style and robust coding that we want to achieve.

ben-l-p: I have a few other branches of cleaned up code lying around, so when I merge those I can do a check and swap everything over. Although I would say it is quite low on the list of things to do at the moment

wong-hl: Agreed, it is very low on the to do list

wong-hl (approved):

kccwing (approved):

kccwing:
Great addition to the code!! As discussed, might be worth extending this to eventually replace the `AsymptoticStability` solver - but again, low on the to do list :)

PR #320, opened 2025-03-23
Hotfix for pypi github action

kccwing:
As titled, also includes update to pyproject.toml for potentially replacing the setup.py build process - tested with [act](https://nektosact.com/) and github action runners.

Issue being addressed is the mayavi direct dependency for pypi uploads - https://stackoverflow.com/questions/54887301/how-can-i-use-git-repos-as-dependencies-for-my-pypi-package


kccwing:
Pending a discussion re #323 this is ready for review, thanks!

PR #325, opened 2025-03-29
Update installation.md

rafapalacios:
Some tiding up of the installation for v2.4

### DIFF
--- a/docs/source/content/installation.md
+++ b/docs/source/content/installation.md
@@ -1,5 +1,5 @@
-# SHARPy v2.3 Installation Guide
-__Last revision 10 June 2024__
+# SHARPy v2.4 Installation Guide
+__Last revision 29 March 2025__
 
 The following step by step tutorial will guide you through the installation process of SHARPy. This is the updated process valid from v2.3.
 
@@ -14,7 +14,7 @@ SHARPy is being developed and tested on the following operating systems:
 * MacOS Mojave and Catalina (Intel)
 * MacOS Sonoma (Apple Silicon M2)
 
-Windows users can also run it by first installing the Windows Subsystem for Linux (https://learn.microsoft.com/en-us/windows/wsl/install) and a XServer such as GWSL, which can be installed through the Microsoft Store. SHARPy is also available to the vast majority of operating systems that are supported by Docker
+Windows users can also run it by first installing the Windows Subsystem for Linux (https://learn.microsoft.com/en-us/windows/wsl/install). It may be also convenient to install a XServer such as GWSL, which are available through the Microsoft Store. SHARPy is also available to the vast majority of operating systems that support Docker.
 
 __Required Distributions__
 
@@ -44,13 +44,14 @@ compile the libraries with custom flags, build it from source (recommended).
 
 ## Building SHARPy from source (release or development builds)
 
-SHARPy can be built from source so that you can get the latest release or (stable) development build.
-
+Building SHARPy from source, ensures that you get the latest release or (stable) development build. 
 SHARPy depends on two external libraries, [xbeam](http://github.com/imperialcollegelondon/xbeam) and
 [UVLM](http://github.com/imperialcollegelondon/UVLM). These are included as submodules to SHARPy and therefore
 once you initialise SHARPy you will also automatically clone the relevant versions of each library.
 
-### Set up the folder structure
+The installation has three steps, cloning, installation, and testing. Different alternatives are included below.
+
+### Step 1: Set up the folder structure
 
 Clone `sharpy` in your desired location, if you agree with the license in `license.txt`.
 ```bash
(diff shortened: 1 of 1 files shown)


docs/source/content/installation.md
```diff
@@ -224,6 +229,43 @@ to your taste. This is compatible with both standalone and Anaconda installation
 
 8. This concludes the installation! Continue reading the [Running SHARPy](#running-sharpy) section.
 
+### Step 2 (PyPI route): Obtain SHARPy from PyPI (experimental!)
```
ben-l-p, line 232: We may want to add some of the disclaimers here about packages which need to be installed prior, as well as it being prebuilt @kccwing ?

ben-l-p: I realise this PR has been sitting for a while... do we want to make any changes about the PyPi side of things @kccwing or are we good to merge?

kccwing: I think @wong-hl is working on the matrix build and hence the tvtk -> vtk to make it happen? Sorry I haven't been keeping up - glad to discuss the next steps.

wong-hl: The matrix build will need a bit of work, as the packaging process needs to be restructured such that each new OS and CPU combination is actually saved as a different wheel. They currently end up in the same wheel and overwrite each other. 

The `tvtk` -> `vtk` is a separate thing. I think @ben-l-p is also working on it??? I've paused working on it because he mentioned he was doing some stuff on that front and the test suite doesn't catch errors

ben-l-p: We should probably get around to merging this after many months... The VTK issues should be all resolved from my previous PR. Is the PyPi still working?

wong-hl: Do you mean the deployment to PyPi or creating different distributions for each os-cpu combination?

ben-l-p: I was meaning the deployment. From what I can see [here](https://pypi.org/project/ic-sharpy/2.4/), it doesn't automatically keep itself up to date with the main GitHub branch, and I'm not overly familiar with how it should work.

wong-hl: The changes here don't affect the CD so assuming that it will use what is on `master` then it will [publish to PyPI when there is a new release](https://github.com/ImperialCollegeLondon/sharpy/blob/98813d7fb746cd90c7c69e2668c89d13f3c7be2d/.github/workflows/pypi_build.yaml#L74)

ben-l-p: Ah makes sense. There aren't many new features since we did the last release, more just fixes, but we could do a new release soon and then all will be well

kccwing: Happy new year both! I'll get back to properly creating the distributions and deploying them, thanks!

wong-hl: Happy new year Kelvin!! I had a look into creating the distributions yesterday as I wanted to play around with [cibuildwheel](https://cibuildwheel.pypa.io/en/stable/ci-services/) on a code base that has native code. I managed to get it working, but it thinks that SHARPy is a pure python wheel... So, I dug a little deeper and I found that when building the `sdist` (source distribution) it was building the native components due to [this line in the `setup.py`](https://github.com/ImperialCollegeLondon/sharpy/blob/98813d7fb746cd90c7c69e2668c89d13f3c7be2d/setup.py#L104) where it invokes the `run()` method to build the native libraries. This meant that the source only distributions and the platform-dependent distributions were identical, thereby making it appear to be a pure python wheel. Thus, the fix for this is to somehow specify in the packaging/build process that there are native components and where they are found.

From a quick dive into the [setuptools docs](https://setuptools.pypa.io/en/latest/index.html) to see how `setup.py` could be modified and skim of the page on [building extension modules](https://setuptools.pypa.io/en/latest/userguide/ext_modules.html), I found that it expects the native code to be a module (e.g. `sharpy.uvlm.sover` for binary that it will load). So, once I saw that it invokes the [shared object directly through ctypes for the ffi](https://github.com/ImperialCollegeLondon/sharpy/blob/98813d7fb746cd90c7c69e2668c89d13f3c7be2d/sharpy/utils/ctypes_utils.py#L18), I gave up and called it a day. It could be that there is a simple fix/trick to it, but it wasn't immediately obvious from the documentation. 

Edit: I also looked at numpy, scikit-learn and pycontrails to see if I could nick some code from them but had no luck

TL;DR - I think creating platform dependent distributions of SHARPy won't be trivial and would take more than a day (depends on how tangled up the build process is and if there is an LLM that knows how to deal with this situation)

ben-l-p: I don't think that it's worth spending too much time on this - the pip install from a git clone is not that troublesome. The only actual need for this is for when I go to Airbus, but I can just repeat the way that Kelvin got it working, even if it isn't the most elegant. If others are happy with this, it might just be a good idea to remove mention of the PyPi route from the docs?

PR #326, opened 2025-05-08
Update linearassembler.py

rmorenoramos:
V2.4 breakes the LinearAeroelastic tests, since the solver is not imported. This pull request restores the working environment for linear simulations by importing the correct solver


ben-l-p:
Thanks for investigating this issue - I will double check that this problem doesn't affect other solvers, and then we can look to merge.

ben-l-p (approved):
Had some time to check through what went wrong; I removed this import in the previous PR because my IDE mistakenly said it was an unused import, and yet the code still passed tests. Thanks for fixing this issue!

kccwing:
btw, test for pypi was cancelled because GitHub deprecated the older ubuntu runner again - this has been addressed in #320 - so feel free to merge this - thanks!

PR #328, opened 2025-09-05
Migrate plotting utilities away from Mayavi/TVTK to VTK

ben-l-p:
Also includes merging main to develop as this should have been done earlier. Some tests are failing with this branch due to changes in the linearised aerodynamic system which are under investigation.

### DIFF
--- a/sharpy/postproc/aerogridplot.py
+++ b/sharpy/postproc/aerogridplot.py
@@ -10,7 +10,7 @@ import sharpy.utils.settings as su
 import sharpy.aero.utils.uvlmlib as uvlmlib
 from sharpy.utils.constants import vortex_radius_def
 
-from ..utils.plotutils import plot_frame_to_vtk
+from sharpy.utils.plotutils import plot_frame_to_vtk
 
 
 @solver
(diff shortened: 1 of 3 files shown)


sharpy/postproc/aerogridplot.py
```diff
@@ -167,205 +168,88 @@ def plot_body(self):
...
-                        '.vtu')
-
-            dims_star = self.data.aero.timestep_info[self.ts].dimensions_star[i_surf, :].copy()
-            dims_star[0] -= self.settings['minus_m_star']
+            filename = f"{self.wake_filename}_{i_surf:02d}_{self.ts:06d}.vtu"
 
-            point_data_dim = (dims_star[0]+1)*(dims_star[1]+1)
-            panel_data_dim = (dims_star[0])*(dims_star[1])
+            zeta_star = np.moveaxis(self.data.aero.timestep_info[self.ts].zeta_star[i_surf], 0, -1).copy() # [m*+1, n+1, 3]
+            gamma = self.data.aero.timestep_info[self.ts].gamma_star[i_surf].copy()  # [m*, n]
```
wong-hl: This is me nitpicking but the comment of `m*+1` is not really clear what it is. The `+1` is what's throwing me off. And should the one comment in the line below be `m*` or just `m`

sharpy/sharpy_main.py
```diff
@@ -4,7 +4,7 @@
 import warnings
 import sys
 import dill as pickle
-import sharpy.utils.cout_utils as cout
+from .utils import cout_utils as cout
```
wong-hl: This has the relative importing as well. There are a few lower down as well

sharpy/postproc/aerogridplot.py
ben-l-p: Will change this, thanks for spotting. In my own codes, I usually denote the shape of arrays with respect to other variables in such comments, but I get why it doesn't make much sense here!

sharpy/sharpy_main.py
wong-hl: Why are you a fan of relative imports? I thought they weren't great and could be ambiguous

sharpy/postproc/aerogridplot.py
wong-hl: Oh! It's `m_star` I thought it was a multiplication. It makes sense now.

wong-hl (approved):

PR #333, opened 2026-04-22
Remove requirement for Mayavi for building SHARPy.

ben-l-p:
Mayavi is now considered un-maintained, and as it has again broken the install process I have depreciated it. All core plotting routines (modal, aero grid and beam) no longer depend upon this and should not be affected. There is still some legacy code referring to it - I have changed this code to use local imports to prevent issues at import time.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -139,9 +139,6 @@ setup(
         "colorama",
         "dill",
         "jupyterlab",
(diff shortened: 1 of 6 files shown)


isabella586 (approved):
Agree with removing the requirement for it. Changes look fine.

PR #335, opened 2026-09-23
Save structured grid VTK output as .vts so Paraview can open it

Aiyo110:
## Bug

Since the move from TVTK to VTK (#331), `AerogridPlot` (body and wake) and the mode shapes written by `Modal` (`write_modes_vtk`) and `AsymptoticStability` (`modes_to_plot`) are written by `plotutils.plot_frame_to_vtk`, which uses `vtkXMLStructuredGridWriter`. The file names kept the `.vtu` extension of the unstructured grids that TVTK used to write.

Paraview chooses the reader from the file extension, and the `.vtu` (unstructured grid) reader refuses a structured grid, so none of these files can be opened in Paraview. With Paraview 6.1.1:

```
File not readable. No reader found for '.../beam_modal_analysis/tstep000000_ModalShape_000002_00.vtu'
```

The `BeamPlot` output (`.vtp`) is not affected.

## Minimal example

On `main`:

```python
import numpy as np
import vtk
from sharpy.structure.utils.modalutils import write_zeta_vtk

x, y = np.meshgrid(np.linspace(0, 1, 3), np.linspace(0, 2, 5), indexing='ij')
zeta = [np.stack((x, y, np.zeros_like(x)))]
write_zeta_vtk(zeta, zeta, 'mode')  # writes mode_00.vtu

print(vtk.vtkXMLUnstructuredGridReader().CanReadFile('mode_00.vtu'))  # 0: the .vtu reader refuses the file
print(vtk.vtkXMLStructuredGridReader().CanReadFile('mode_00.vtu'))    # 1: the file is a structured grid
```

Any case with `AerogridPlot`, or with `write_modes_vtk` in `Modal`, shows the same problem, e.g. `tests/linear/goland_wing/test_goland_flutter.py`.

## Fix

- Name these files `.vts`, the extension VTK uses for XML structured grids (`vtkXMLStructuredGridWriter().GetDefaultFileExtension()`). The contents of the files are unchanged: apart from the name, they are byte-identical to the files written on `main`.
- Update the docstrings of `plot_frame_to_vtk`, `write_zeta_vtk` and `AsymptoticStability`, and the description of the output files in `docs/source/content/installation.md` (which also still said that the beam is written as `.vtu`; it is `.vtp`).
- Ignore `*.vts` in `.gitignore`, as `*.vtu` already is.

`AerogridPlot.plot_nonlifting_surfaces` is not changed: it still writes an unstructured grid, so `.vtu` is correct there.

Scripts that look for this output as `*.vtu` will need to look for `*.vts` instead.

## Tests

New `tests/utils/test_vtk_file_extensions.py`:

- `test_write_zeta_vtk` writes a small lattice and checks that it is saved as `.vts` and reads back with the expected points, cells and displacement.
- `test_case_output` runs a small Goland case (`BeamLoader`, `AerogridLoader`, `AerogridPlot`, `BeamPlot`, `Modal` with `write_modes_vtk`) and checks that every VTK file it writes can be read by the reader associated with its extension.

Both tests fail on `main` and pass with this change, and `tests/utils` and `tests/linear/goland_wing` pass. I also ran the Goland flutter case with `modes_to_plot = [0, 1]` and opened every VTK file it writes in Paraview 6.1.1: on `main` only 1 of the 63 files opens (the beam `.vtp`); with this change all 63 do.

Tested on macOS (Apple Silicon), Python 3.11, VTK 9.6.2, Paraview 6.1.1.


Aiyo110:
The failing `build (3.10.8)` check is not caused by this change. All 35 errors are the same `TypeError: SetPoints argument 1: method requires a VTK object`, raised at `sharpy/postproc/beamplot.py:259` (`vtk.vtkPolyData(points=coords)`), which this PR does not touch. The CI installed VTK 9.7.0 (released 15 August), and with VTK 9.7.0 `vtkPolyData(points=<numpy array>)` raises this error, while VTK 9.6.2 accepts it. The last test run on `main` (7 May) predates VTK 9.7.0, so `main` would fail in the same way now. `test_write_zeta_vtk` passes in this run; `test_case_output` fails only because it runs `BeamPlot`. I can open a separate PR that makes `BeamPlot` work with VTK 9.7.

